PR #4, opened 2019-11-16
Icon component

Joozty:
@viktor-yakubiv This is a kickoff. I would really like to hear your opinion about this setup. I added auto-generated docs as you may see [here](https://design-git-init-react.oacore.now.sh/). It's just a showcase. 

More text is coming soon. Unfortunately SVG sprites are tricky to make work. I decided to not use them at all.


Joozty:
> It's a possibility to write less with a couple of strings for automation.

Write less? I don't see anything which saves lines of code.

I wasn't clear. I meant it that it's not hard to generate one big sprite for all icons. The hardest part is to generate a new sprite from that big one sprite in another project. Currently, there is no sprite generated but I can change it if you insist. I still don't see any big differences in whether icons are loaded via JS code (in one chunk) or in a sprite. Even I think from the point of speed the SVG icons in JS code are faster because webpack packs only needed icons instead of the whole sprite. On the other hand, that code is not cached by the browser.

viktor-yakubiv:
The point is that:

1. You have ability easily create colour transitions for any icon
2. You can use any icon from the package without explicit importing into JS
3. You have as simple as possible Icon component

Also, I assume that we will not need extra icons per project. Icon library should be universal and cross-project. However, if this hypothesis is wrong we can sort it out in the future.

Joozty:
@viktor-yakubiv Fair enough let me slightly change the PR to add sprite generation :)

PR #11, opened 2019-11-18
Switch to PostCSS & enable CSS modules

viktor-yakubiv:
Replaces SASS/SCSS with pure CSS with help of PostCSS and PostCSS Preset Env.

Adds `postcss-modules` for CSS building and `css-modules-transform` for including CSS class names into JS.

After merging this, we will require to include CSS files separately into the dependent projects.


viktor-yakubiv:
I added an index file for CSS but this is not transformed properly. We will need to add `postcss-import` later.

src/components/icon/index.jsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react'
-import './index.scss'
 import PropTypes from 'prop-types'
+
+import { icon as iconClassName } from './index.css'
```
Joozty: This is undefined  😢

Joozty: when you use it like this it works just fine.

```
import css from './index.css'

...

    <svg className={css.icon} viewBox={`${icon.viewBox}`}>
      <use xlinkHref={`#${icon.id}`} />
    </svg>
```

src/components/icon/index.css
```diff
@@ -0,0 +1,6 @@
+.icon {
```
Joozty: When I run `npm run build:css` it generates file like this. The class name has suffix weird hash. Will be this hash different from build to build? What if I want to change the icon color in let's say Reader? 

```
.index-icon-3kEAc {
  display: inline-block;
  width: 1.5rem;
  height: 1.5rem;
  fill: red;
}
```

webpack.config.js
```diff
@@ -23,14 +23,11 @@ module.exports = {
...
-        test: /\.s[ac]ss$/i,
+        test: /\.css$/,
         use: [
-          // Creates `style` nodes from JS strings
           'style-loader',
-          // Translates CSS into CommonJS
-          'css-loader',
-          // Compiles Sass to CSS
-          'sass-loader',
+          { loader: 'css-loader', options: { importLoaders: 1 } },
```
Joozty: Wouldn't it be better to extract CSS here instead of using `npm run build:css`? 

```
        test: /\.css$/,
        use: [
          DEV ? 'style-loader' : MiniCssExtractPlugin.loader,
          { loader: 'css-loader', options: { importLoaders: 1 } },
          'postcss-loader',
        ],
```

src/components/icon/index.jsx
viktor-yakubiv: It's weird. I read the Babel output, it was correct. I will take a look

src/components/icon/index.css
viktor-yakubiv: If you go below, you fill is `currentColor`. This is a special keyword that means the fill inherits current element `color` property that is inherited from parent element.

webpack.config.js
viktor-yakubiv: Not really. It's a library. Even after build a developer should be able to read source. Minifier always set up on the end projects.

src/components/icon/index.css
Joozty: I meant it like it would be weird to use this selector `.index-icon-3kEAc` for changing any property of icon in Reader. Color was wrong example. Let's say I would want to change width.

viktor-yakubiv: You pass another class name. It will be supported later. I am considering a way but I feel I do not have any other [option](https://github.com/JedWatson/classnames).

viktor-yakubiv: To answer your question better, hash is base64-encoded selector content, so it's the same if style did not change and different otherwise. It's added to make a class name unique.

Joozty: If the only option is to pass another class name then I am strongly against this approach as I don't see any additional value. It will be just a big mess in imported packages. If we want to have own CSS framework we need to have some helper classes. For example `d-flex`. I won't be able to use it in for example in Reader unless I write something like `d-flex-dhdjf23` which is a mess.

viktor-yakubiv: Please, take your time to read the [docs](https://github.com/css-modules/css-modules).

I've chosen CSS modules because I could not easily contribute to the Reader because of class name conflict. So, I decided that CSS file with ugly class names on production is not that bad.

viktor-yakubiv: I was exactly like you 2 years ago, by the way 🙂

Joozty: I have not read the documentation yet. But it doesn't change my mind regarding unique names. We are building reusable library similar as bootstrap and it doesn't use base64 suffixes too. I can see use case with unique classnames in final product but not in reusable library. if the classnames conflicts are the reason then add just prefixes to all classes like `.deisgn-icon`

webpack.config.js
Joozty: AFAIK it's not minifier. it should do just the work of `npm run build:css` for you automatically. 

> This plugin extracts CSS into separate files. It creates a CSS file per JS file which contains CSS. It supports On-Demand-Loading of CSS and SourceMaps.

viktor-yakubiv: I do not completely understand this. As far as I know we use Webpack only because of react-stylguide and for development server. However, the production build goes with simple NPM scripts.

To extend, I am against building with Webpack so I do not mind this only for docs.

I payed attention to `MiniCssExtractPlugin.loader` only, sorry. I think we do not need this. However, if talk about style loader, it's good for some cases but not here because we really need it separate.

My idea was to allow user do the following:
```js
import { Icon, Button } from '@oacore/design'
import '@oacore/design/index.css'
```
along with
```js
import Icon from '@oacore/design/icon'
import '@oacore/design/icon/index.css'
```

In the first case a user imports whole library (because he knows that he is using whole library) with all the css in a single file. In the second case the user imports only _Icon_ with its CSS only because this is all he/she needs. To give an ability to choose I generate many CSS/JS files and I think this is good approach.

Although, I personally do not really like how it look now so we have to explore it more. Do you have any opinion?

Joozty: Well, I thought we would use webpack for the building process too. We need to transpile the code and build one module from it. Currently, it is done with `npm run start` script. 

We have the same idea the only thing I am not sure about is whether this is possible 
`import Icon from '@oacore/design/icon'` I am not sure how to do it. The only solution I am aware of is creating sub-packages




.

src/components/icon/index.css
viktor-yakubiv: It's not CSS framework. It's a component library. The difference is that you ideally do not have to care about CSS, just take a component and use it.

If you mind, we can switch CSS modules off. However, we will have mess when switch it on back. So I would no like to do it.

webpack.config.js
viktor-yakubiv: Not really. It's like `import Link from 'next/link'`. Link is not a sub-package. It's possible to resolve package subdirectories but we have to find a way to publish it right.

viktor-yakubiv: About Webpack, we do not have to. We do not need to use Webpack twice. Webpack on the final project will resolve all dependencies and bundle everything once. We just need to transpile JS and CSS because Babel and other tools usually skip `node_modules`.

src/components/icon/index.css
Joozty: I don't want to switch off css modules I just want you to remove suffixes. That's all. 

IMHO we should not design only components but also create helpers classes and layout too. We can borrow bootstrap or anything else for the start I don't mind. The idea is to have as much as possible of css code in this library and minumum css code in packages that will use this library.

viktor-yakubiv: I agree but I am agains helpers kind of `.d-flex`. People often overuse it. You do not have to build components from helper classes. That is why I am against it.

However, I am not against helpers kind of `.h1`.

webpack.config.js
Joozty: But all the libraries used here will need to be peer dependencies in other packages, won't be? The ideal scenario would be to have only two peer dependencies react and react-dom. Everything else should be installed here. Maybe I am getting it wrong. If you know what you are doing then go for it and try to make a minimal example for it.

src/components/icon/index.css
viktor-yakubiv: > want you to remove suffixes

Removing suffixes means switching modules off. If we do so, we could not change the name of file, because it's used in final class name. Suffix if more about making class name weird to tell the user that he overrides it on his own risk because this might change.

It's basically taking ability to be the only one who can control class name. Others can only [extend the component but cannot change](https://en.wikipedia.org/wiki/Open–closed_principle).

webpack.config.js
viktor-yakubiv: No, `dependencies` are always installed if it's listed but `devDependencies` skipped. `peerDependencies` throw warning if absent.

src/components/icon/index.css
Joozty: Your call then. The only thing I would want to see are helper classes without prefixes and suffixes at least.

webpack.config.js
viktor-yakubiv: I could not find a way to avoid `dist` in module path. There are a few options, I created #14 to discuss this.

viktor-yakubiv:
Thank you but please do not merge unless you are an assignee.

PR #15, opened 2019-11-19
Generate SVG sprites. Improve icon component

viktor-yakubiv:
It makes Icon component less smart but more convenient (and dangerous as well).

Along with it:
1. Adds processing SVG images with SVGO
2. Moves assets to `src/assets`
3. Adds building SVG-sprites with `svgstore-cli`
4. Simplifies integration with Now (#17)
5. Fixes (or truing to fix) npm-install warnings


package.json
```diff
@@ -59,7 +59,8 @@
     "build:js": "babel src --out-dir dist --verbose",
     "build:css": "postcss src/**/*.css --base src --dir dist",
     "build:docs": "styleguidist build",
-    "build": "npm run build:css && npm run build:js && npm run build:docs",
+    "build:svg": "svgo -f src -o dist -r --disable=removeViewBox --enable=removeDimensions",
+    "build": "npm run build:svg && npm run build:css && npm run build:js && npm run build:docs",
```
Joozty: Maybe worth of running it in parallel

```
"build": "npm-run-all build:*",
```

src/components/icon/index.jsx
```diff
@@ -3,51 +3,29 @@ import PropTypes from 'prop-types'
...
-    <svg className={iconClassName} viewBox={`${icon.viewBox}`}>
-      <use xlinkHref={`#${icon.id}`} />
-    </svg>
-  )
-})
+/**
+ * The component creates an inline SVG image based on
+ * `src` and `alt` attributes.
+ */
+const Icon = ({ src, alt, ...restProps }) => (
```
Joozty: I am not sure about passing src with the whole path. On the other hand we can have some "postInstall" script which copies the assets folder to `public` folder in next based projects.

src/components/icon/index.jsx
```diff
@@ -3,51 +3,29 @@ import PropTypes from 'prop-types'
 
 import { icon as iconClassName } from './index.css'
 
-import DownloadIcon from './assets/download.svg'
-import OutlineIcon from './assets/outline.svg'
-import ThumbnailsIcon from './assets/thumbnails.svg'
-import PaperInfoIcon from './assets/paper_info.svg'
-import PrintIcon from './assets/print.svg'
```
Joozty: It's a matter of taste but I like to have distinguished files. When new icons are added it will be hard to find index.js file. Just a note. No need to change it. 🙂

package.json
viktor-yakubiv: Not everything may be ran in parallel now but I will add it. Thank you!

src/components/icon/index.jsx
viktor-yakubiv: It's more about to start with something extremely simple an see how it works for us

src/components/icon/index.jsx
viktor-yakubiv: It's better to have less changes. If you add icon, you need only to place a file and everything works. If you have imports, you have to add a file and update JS. It may have significant impact of mistake at some cases.

Joozty: Yeah I agree

Joozty:
@viktor-yakubiv svg icon is not displaying in docs.

missing_icons

viktor-yakubiv:
@Joozty I know. This is because React Styleguidist is not properly configured so CSS classes are not added despite being loaded. I think like we have to add `css-modules` loader to web pack config unless we have one.

You said you are going to improve Webpack and Styleguidist configs. I hope this is still on you.

viktor-yakubiv:
Thank you but please do not merge unless you are an assignee.

Joozty:
Sorry I am used to a process when review approval is given PR can be merged unless it's ready or there is a label `work-in-progress` or `do-not-merge`

PR #18, opened 2019-11-20
Setup CSS environment

viktor-yakubiv:
1. Improves PostCSS configuration
2. Adds Stylelint


.prettierrc.yml
```diff
@@ -1,7 +1,3 @@
 trailingComma: es5
 semi: false
 singleQuote: true
-overrides:
```
Joozty: We should migrate to @oacore/configs as soon as possible. I am still waiting for your review of config and then I will create pull request

.stylelintrc.yml
```diff
@@ -0,0 +1,10 @@
+extends:
```
Joozty: Maybe worth of migrating to @oacore/configs, what do you think?

Joozty: Obviously I can handle that, you don't need to spending time on this only if you want to. 🙂

.prettierrc.yml
viktor-yakubiv: I know. This is the only place we have Stylelint config. Let's start from this place and see how it works. I do not want often update @oacore/configs because something does not fit our workflow. It means many releases with big loop.

.stylelintrc.yml
viktor-yakubiv: See above

.prettierrc.yml
Joozty: This one is prettier config. We can postpone moving stylelint. I would definitely like to see it in configs one day for sure. :)

Joozty: However I still insist on using prettier config from @oacore/configs.

viktor-yakubiv: Yes, I missed this. Sure. Feel free to push this update. I did not even notice that prettier config is local.

viktor-yakubiv: I mean, we can replace it later

PR #20, opened 2019-11-20
Bootstrap Reboot and typography

viktor-yakubiv:
Copied and pasted basics from Bootstrap to normalise page styles.

### DIFF
--- a/src/foundation/reboot.css
+++ b/src/foundation/reboot.css
@@ -1,0 +1,585 @@
+// stylelint-disable at-rule-no-vendor-prefix, declaration-no-important, selector-no-qualifying-type, property-no-vendor-prefix
+
+/* Reboot
+ *
+ * Bootstrap Reboot
+ *
+ * Normalization of styles, manually forked from Bootstrap.
+ * Bootstrap is licensed MIT. https://github.com/twbs/bootstrap
+ *
+ * Normalization of HTML elements, manually forked from Normalize.css to remove
+ * styles targeting irrelevant browsers while applying new styles.
+ * Normalize is licensed MIT. https://github.com/necolas/normalize.css
+ *
+ */
+
+// Document
+//
+// Change from `box-sizing: content-box` so that `width`
+// is not affected by `padding` or `border`.
+*,
+*::before,
+*::after {
+  box-sizing: border-box;
+}
+
+// Root
+//
+// 1. Ability to the value of the root font sizes, affecting the value of `rem`.
+//    null by default, thus nothing is generated.
+
+:root {
+  font-size: var(--font-size-root); // 1
+}
+
+// Body
+//
+// 1. Remove the margin in all browsers.
+// 2. As a best practice, apply a default `background-color`.
+// 3. Prevent adjustments of font size after orientation changes in IE on
+//    Windows Phone and in iOS.
+// 4. Change the default tap highlight to be completely transparent in iOS.
+body {
+  margin: 0; // 1
+  font-family: var(--font-family-base);
+  font-size: var(--font-size-base);
+  font-weight: var(--font-weight-base);
+  line-height: var(--line-height-base);
+  color: var(--body-color);
+  text-align: var(--body-text-align);
+  background-color: var(--body-bg); // 2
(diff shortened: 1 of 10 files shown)


viktor-yakubiv:
@Joozty this is not finished yet but is ready for the review.

Unfortunately, I cannot make it work well without workarounds and conventions so I decided to postpone this kind of decision. So, currently the system leaves `@extend` rules in the final build if it's not resolved. I hope if we need, we can sort it out later but it worth to create an issue for it.

The only thing I have to finish, change custom properties' values.

src/foundation/variables.css
```diff
@@ -1 +1,180 @@
...
+  --legend-font-weight: null;
+  --mark-padding: 0.2em;
+  --mark-bg: #fcf8e3;
+  --dt-font-weight: var(--font-weight-bold);
+  --nested-kbd-font-weight: var(--font-weight-bold);
+  --list-inline-padding: 0.5rem;
+
+  /* Code */
+  --code-font-size: var(--small-font-size);
+  --code-color: var(--pink);
```
Joozty: It doesn't exist too.

src/foundation/reboot.css
```diff
@@ -0,0 +1,585 @@
...
+// 1. Duplicate behavior to the data-* attribute for our tooltip plugin
+// 2. Add the correct text decoration in Chrome, Edge, IE, Opera, and Safari.
+// 3. Add explicit cursor to indicate changed behavior.
+// 4. Prevent the text-decoration to be skipped.
+
+abbr[title],
+abbr[data-original-title] {
+  // 1
+  text-decoration: underline; // 2
+  text-decoration: underline dotted; // 2
```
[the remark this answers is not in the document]
viktor-yakubiv: If you see more than one CSS property it's always because of browser support. All browsers apply the first one. However, second one will be applied only if browser understands that.

src/foundation/variables.css
viktor-yakubiv: Thank you for pointing this out. I will add colour schema and missing variables.

viktor-yakubiv: By the way, [var()](https://developer.mozilla.org/en-US/docs/Web/CSS/var) accepts default value as default value if custom property does not exist.

Should I provide default values everywhere, should not I?

Joozty: I would rather define those variables or not use var at all.

src/foundation/variables.css
```diff
@@ -1 +1,180 @@
...
+  --rounded-pill: 50rem;
+  --box-shadow-sm: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.075);
+  --box-shadow: 0 0.5rem 1rem rgba(0, 0, 0, 0.15);
+  --box-shadow-lg: 0 1rem 3rem rgba(0, 0, 0, 0.175);
+  --box-shadow-inset: inset 0 1px 2px rgba(0, 0, 0, 0.075);
+  --component-active-color: var(--white);
+  --component-active-bg: var(--primary);
+  --caret-width: 0.3em;
+  --caret-vertical-align: calc(var(--caret-width) * 0.85);
+  --caret-spacing: calc(var(--caret-width) * 0.85);
```
viktor-yakubiv: @Joozty I am in order to remove this but I am not really sure about.

What do you think about [Bootstrap carets](https://github.com/twbs/bootstrap/blob/b2ae6a6a94a70a027469fa64723e8268ca66254b/scss/mixins/_caret.scss)? Is it good idea?

They use old-fashion border trick when you have 2 transparent borders and 1 coloured, it draws a triangle. It's a zero-dependency carets what I like but on the other hand it's little bit tricky what is confusing me.

I think I will remove and add it back when needed. We will need to support mixing for this feature.

Joozty: Yesterday I discovered how those carets work. It took me some time to understand it. 😄 

I this trick [in dashboard search bar](https://github.com/oacore/dashboard/blob/layout/components/searchbar/index.scss#L10-L20). I guess it's of having it in the design library. However, we can postpone it. :)

viktor-yakubiv:
Closing in case of #32

PR #34, opened 2019-12-02
Use @oacore/configs

Joozty:
@viktor-yakubiv I hope in the next major version we will have stylelint config too. 🙂


.eslintrc.yml
```diff
@@ -1,37 +1,11 @@
 extends:
-  - airbnb
-  - prettier
-  - prettier/react
+  - '@oacore/eslint-config-react'
 
+parser: babel-eslint
```
viktor-yakubiv: I wonder if we can add a parser to `@oacore/eslint-config-react` too

PR #35, opened 2019-12-02
Add headings documentation

viktor-yakubiv:
Adds `Heading` component and a simple documentation about how CSS custom properties work with the typography.

### DIFF
--- a/src/foundation/README.md
+++ b/src/foundation/README.md
@@ -1,0 +1,44 @@
+## Headings
+
+### Application headings
+
+This set of headings are useful when you need it for applications,
+where content is dense and you basically need small headings.
+
+```jsx
+<h1>Heading 1</h1>
+<h2>Heading 2</h2>
+<h3>Heading 3</h3>
+<h4>Heading 4</h4>
+<h5>Heading 5</h5>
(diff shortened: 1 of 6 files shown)


viktor-yakubiv:
I could not manage to use generic styles from `:root` and `body` so I do not really like how the headings look like. However, works everything except the font-family.

src/foundation/heading.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+const Heading = ({
+  children,
+  className,
+  level,
+  display,
+  tag: Tag = `h${level}`,
+  ...restProps
+}) => (
+  <Tag className={`${display ? 'display' : ''} ${className}`} {...restProps}>
```
Joozty: Would be nice to start using `classNames`

viktor-yakubiv: I would add it later

src/foundation/README.md
```diff
@@ -0,0 +1,44 @@
+## Headings
+
+### Application headings
+
+These set of headings are useful when you need it for applications,
```
Joozty: **This** set of headings 🙂

PR #37, opened 2019-12-03
Change build directory to lib

viktor-yakubiv:
I am afraid I missed something in some config file. @Joozty could you please check?


viktor-yakubiv:
Please, be aware that `dist` is not ignored anymore so if you have old build you may accidentally push it to the repository.

PR #38, opened 2019-12-04
Fix components font family in docs

viktor-yakubiv:


styleguide.config.js
```diff
@@ -33,6 +33,28 @@ module.exports = {
     },
   },
 
+  styles: {
```
Joozty: This should work too. Maybe the config would be more readable.


```
    styles: {...},
	require: [
		'path/to/styleguide.css',
	],
```


or this one https://react-styleguidist.js.org/docs/cookbook.html#how-to-change-styles-of-a-style-guide

viktor-yakubiv: I required all our CSS but fond does not work, so I had to copy and paste it here.

PR #39, opened 2019-12-04
Add TextFields

viktor-yakubiv:
Kind of works for the start.


Joozty:
@viktor-yakubiv Is it ready or it's still a draft?

src/elements/forms/text-field.jsx
```diff
@@ -0,0 +1,64 @@
...
+  ...inputProps
+}) => {
+  const controlId = [id, 'control'].join('-')
+
+  return (
+    <Tag
+      id={id}
+      className={classNames
+        // prevents passing 'default' to the class list
+        .use('container', { [size]: size !== 'default' }, className)
```
Joozty: There can be a collision. When className prop contains for example `container` or whatever class defined in `text-fields.css`. I think you are aware of it but would be nice to keep it in mind and solve it somehow. Not necessary here and now. 

```jsx
<TextField
  id="login-password"
  type="password"
  name="password"
  label="Password"
  className="container"
  placeholder="🔒 Something secure"
/>
```

Current output:

```
<div class="text-field-container-1W0Hy text-field-container-1W0Hy">...</div>
```

Expected output:
```
<div class="text-field-container-1W0Hy container">...</div>
```

src/elements/forms/text-field.css
```diff
@@ -0,0 +1,76 @@
...
+
+input {
+  display: block;
+  width: 100%;
+  padding: var(--form-control-padding-y) var(--form-control-padding-x);
+  border: 1px solid var(--primary);
+  border-radius: var(--form-control-corner-radius, 2px);
+  height: calc(2 * var(--form-control-padding-y) + 1em);
+
+  &:focus {
```
Joozty: Didn't you want to avoid this syntax?

Joozty: Didn't you want to avoid this syntax?

src/elements/forms/text-field.jsx
```diff
@@ -0,0 +1,64 @@
...
+        .withModule(textFieldClassNames)}
+    >
+      <input id={controlId} {...inputProps} />
+      <label htmlFor={controlId}>{label}</label>
```
Joozty: It would be nice to render this element only when the label prop is provided.

src/elements/forms/text-field.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import textFieldClassNames from './text-field.css'
+
+import { classNames } from 'utils'
+
+const generateId = () =>
+  // Math.random should be unique because of its seeding algorithm.
```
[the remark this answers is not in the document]
viktor-yakubiv: Unique IDs is not a problem for JS. There are many libraries on NPM that who'll solve this. However, I did not want to have another dependency for such small stuff that I might remove or replace later. I did this just in case and used a gist you linked.

So, you are right, I have to credit it from the code.

src/elements/forms/text-field.jsx
viktor-yakubiv: I am aware bu I cannot find convenient solution. The best what I can find is either

```js
classNames.use(...classes).from(module).add(customClasses)
```

or process chain differently, i.e. do not do perform it in `toString()` but do it right in the method.

Any suggestions?

src/elements/forms/text-field.css
viktor-yakubiv: No, this particular syntax is useful. I wanted to avoid nesting that constructs additional classes such as `&__child` because it might confuse and makes searching tricky. In this case it's clear.

src/elements/forms/text-field.jsx
viktor-yakubiv: No, not really. It must always exist. It may be for screen-readers only but I do not know how to pass a parameter for it. Any suggestions?

Joozty: Do you mean like parameter for screen-readers only?

viktor-yakubiv: Yes. This is the only need.

Also, if you want an input element without label, you can just use `<input>` instead.

src/elements/forms/text-field.jsx
Joozty: I don't know. I think you can leave it as is till we run into this problem. You probably won't like it since it's YAGNI principle I guess. If you want to fix it then I vote for the first solution.

viktor-yakubiv: Not really in this case. Could you please create an issue for this?

src/elements/forms/text-field.jsx
Joozty: I would simply copy-paste .sr-only class from bootstrap and add a new prop to TextField component like `labelsronly=false` which would conditionally add this class to label element.

PR #40, opened 2019-12-05
Enable linting *.config.js files

viktor-yakubiv:
1. Adds overrides section to ESLint config
2. Renames Prettier config to make it lintable
3. Fixes all config files after linting

### DIFF
--- a/src/foundation/colors.css
+++ b/src/foundation/colors.css
@@ -41,8 +41,8 @@
 
   /* Success */
   --success: #8bc34a;
-  --success-light: ;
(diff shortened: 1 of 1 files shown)


.eslintrc.yml
```diff
@@ -9,3 +9,11 @@ settings:
     node:
       paths: ['./src']
       extensions: ['.js', '.jsx']
+
+overrides:
```
Joozty: Maybe you could move this section to @oacore/configs and release it. And after that finish this PR.

PR #46, opened 2019-12-11
Add basic buttons

viktor-yakubiv:


viktor-yakubiv:
@Joozty this does not work for this moment. I have to fix pseudo classes.

Also, I need a parameter that describes a type of [button](https://material.io/components/buttons/), that can be either text, outlined or contained. I cannot use `type` prop because it can be passed to a `<button>` HTML element. `color` does not fit current case, `kind` I just do not like. Any suggestions?

Joozty:
@viktor-yakubiv What about `variant`?

src/elements/button/button.css
```diff
@@ -0,0 +1,104 @@
...
+  @extend %button;
+}
+
+.button {
+  @extend %button;
+}
+
+/* Variants */
+
+.text {
```
Joozty: This empty class makes me think about whether it's necessary here. I guess it's good for keeping consistency.

postcss.config.js
```diff
@@ -13,14 +13,19 @@ module.exports = {
     calc(),
     presetEnv({
       stage: 0,
+      features: {
+        'color-mod-function': {
+          unresolved: 'ignore',
```
Joozty: Is it good to ignore it?

src/elements/button/button.css
```diff
@@ -0,0 +1,104 @@
...
+    &:focus {
+      box-shadow: var(--button-pressed-shadow), var(--button-focus-ring);
+    }
+  }
+}
+
+button,
+[type='button'],
+[type='submit'],
+[type='reset'] {
```
Joozty: Interesting didn't know about this type.

src/elements/button/button.css
```diff
@@ -0,0 +1,104 @@
...
+  --button-background: var(--primary);
+  --button-color: var(--white);
+  box-shadow: 0 1px 2px #0003;
+
+  &:hover {
+    box-shadow: 0 1px 2px #0003;
+  }
+
+  &:focus {
+    box-shadow: 0 1px 2px #0003, var(--button-focus-ring);
```
Joozty: Consider creating a variable for this color.

postcss.config.js
viktor-yakubiv: I do not know but I used this function only once for this moment.

Joozty: I would prefer `warn` level at least.

src/elements/button/button.css
viktor-yakubiv: I think more about having shadow variables. Colours are also good but shadows are more important. Let's think about it later.

PR #47, opened 2019-12-12
Add labels

viktor-yakubiv:
Add as simple as possible labels that can be used to highlight OAI in data tables.

### DIFF
--- a/src/elements/label.jsx
+++ b/src/elements/label.jsx
@@ -1,0 +1,33 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import labelStyles from './label.css'
+
+import { classNames } from 'utils'
+
+const Label = ({
+  children,
+  color,
+  className,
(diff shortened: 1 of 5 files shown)


src/elements/label.jsx
```diff
@@ -0,0 +1,33 @@
...
+      .use('label', color !== 'default' && color, className)
+      .withModule(labelStyles)}
+    {...restProps}
+  >
+    {children}
+  </Tag>
+)
+
+Label.propTypes = {
+  color: PropTypes.oneOf(['default', 'primary', 'success', 'danger']),
```
Joozty: You forgot to add a tag here. (it's missing in heading too)

label

viktor-yakubiv: I want to somehow ignore this generic prop globally but there is no ignore list. I know how to do it and will do it later. It's not a priority.

You know that the tag is always a React node.

PR #48, opened 2019-12-12
Add tables

viktor-yakubiv:
Adds table components. Quite a wide range.

Adds tables styles.


viktor-yakubiv:
@Joozty this adds styled tables but the interface is not really good. I would be happy to get your feedback. Perhaps, it's better to remove useless components for time being.

Joozty:
@viktor-yakubiv Let me think about this one if you don't mind. I may need maybe slightly different components. (https://github.com/oacore/dashboard/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc)

src/elements/tables/body.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+
+/**
+ * Empty component for the future purposes.
+ */
+const Body = ({ children, tag: Tag = 'tbody', ...restProps }) => (
```
Joozty: Could you please wrap it into React.useRef(). I want to observe this element. Actually would be nice to wrap all components.

Joozty:
Also, I would appreciate having one component more, which would show the order direction of whole column. Something like this:

<HeaderCell desc={true} asc={false}>Title</HeaderCell>

src/elements/tables/body.jsx
viktor-yakubiv: I suggest you to use pure HTML elements. Maybe it's better than complex API I created.

I should remove these components I guess. Although, [_HeaderCell_](https://github.com/oacore/design/pull/48#issuecomment-565966268) is good idea and we should have it.

Joozty: I like it and it's not that complex. It works just like a proxy. Let's keep it, please.

Joozty: Also, text inside `HeaderCell` should be bigger and bold than in `Cell`

Joozty:
Sorry, I forgot one more thing. Maybe you could consider to export components like:

```
<Table.Row />, <Table.Cell />, ...
``

File: src/elements/tables/body.jsx (hunk 1 above)
Comment (viktor-yakubiv) [reply src/elements/tables/body.jsx]: Why?

Comment (Joozty) [reply src/elements/tables/body.jsx]: Well because it's how you designed it. I mean the font style can be applied on TableHead instead.
h

Comment (viktor-yakubiv) [reply src/elements/tables/body.jsx]: I might missed something. Thank you for pointing this up

Comment (viktor-yakubiv) [reply src/elements/tables/body.jsx]: It should be 1rem (16px) everywhere for this moment. We will see how it works in the future.

File: src/elements/tables/index.js [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+Table.Cell = Cell
+
+export default Table
+export {
+  Table,
+  Head as TableHead,
+  HeadCell as TableHeadCell,
+  Body as TableBody,
+  Row as TableRow,
+  Cell as TableCell,
```
Joozty: These elements shouldn't be exported, should they be?

src/elements/tables/table.md
```diff
@@ -0,0 +1,29 @@
+```jsx
+import { TableHead, TableRow, TableHeadCell, TableCell } from '.';
+
+<Table>
+  <TableHead>
```
Joozty: It would be nice to use a new markup, i.e. `<Table.Row />`

src/elements/tables/index.js
viktor-yakubiv: Why not. You used this components in such way now

src/elements/tables/table.md
viktor-yakubiv: It's not a big deal now. You know that you can use it in such way. For this moment this is enough 😉

Issue #49, opened 2019-12-12
Add CORE logo


viktor-yakubiv:
Add CORE logo into the library.

## Tasks

- [ ] Design CORE logo according to the Material Design guidelines for icons
- [ ] Write a component

PR #52, opened 2019-12-13
Add AppBar

### LINKED ISSUE #44: className utility
There can be a collision in class names. Let's  have a text field component as an example:

When className prop contains for example `container` or whatever class defined in `text-fields.css`. It recognizes it as class name from the module.

```jsx
<TextField
  id="login-password"
  type="password"
  name="password"
  label="Password"
  className="container"
  placeholder="🔒 Something secure"
/>
```

Current output:

```
<div class="text-field-container-1W0Hy text-field-container-1W0Hy">...</div>
```

Expected output:
```
<div class="text-field-container-1W0Hy container">...</div>
```

viktor-yakubiv:
1. Fixes classNames utility, closes #44 
2. Fixes links specificity to have less `!important`s
3. Adds simple AppBar component

### DIFF
--- a/src/elements/app-bar/brand.jsx
+++ b/src/elements/app-bar/brand.jsx
@@ -1,0 +1,19 @@
+import React from 'react'
+
+import styles from './styles.css'
+
+import { classNames } from 'utils'
+
+const Brand = ({ children, className, tag: Tag = 'a', ...restProps }) => (
+  <Tag
+    className={classNames
+      .use('brand')
+      .from(styles)
+      .join(className)}
+    {...restProps}
+  >
+    {children}
+  </Tag>
+)
+
+export default Brand
--- a/src/elements/app-bar/styles.css
+++ b/src/elements/app-bar/styles.css
@@ -15,17 +15,29 @@
   box-shadow: 0 1px 2px #00000042, 0 4px 8px #00000021;
 }
 
+/* Modifiers */
+
 .fixed {
   position: fixed;
 }
 
-.brand {
+/* Children */
+
+.item {
   display: inline-block;
   padding: var(--app-bar-padding);
   font-weight: var(--app-bar-brand-font-weight, 600);
   color: inherit;
   text-decoration: none;
 
+  /**
+   * We should find a generic way to work around buttons too.
+   *
+   * There are not many options:
+   * 1. Use :is(button) for button element and [type=button] and similar
+   *    for other button-like elements
+   * 2. Use some workaround like .interactive
+   */
   &:link:hover,
   &:visited:hover,
   &:focus {
@@ -34,4 +46,8 @@
     background: var(--app-bar-hover-background);
     outline: none; /* May harm user experience */
   }
+}
+
+.brand {
+  /* Is completely nested from item. Kept for the future */
 }
--- a/src/elements/app-bar/styles.css
+++ b/src/elements/app-bar/styles.css
@@ -1,0 +1,37 @@
+:root {
+  --app-bar-padding-x: var(--component-padding-x, 1rem);
+  --app-bar-padding-y: var(--component-padding-y, 1rem);
+  --app-bar-padding: var(--app-bar-padding-y) var(--app-bar-padding-x);
+  --app-bar-background: var(--white, #fff);
+  --app-bar-color: var(--gray-800, inherit);
+  --app-bar-hover-background: #0000000a;
+}
+
+.container {
+  display: flex;
+  width: 100%;
+  color: var(--app-bar-color);
+  background: var(--app-bar-background);
+  box-shadow: 0 1px 2px #00000042, 0 4px 8px #00000021;
+}
+
+.fixed {
+  position: fixed;
+}
+
+.brand {
+  display: inline-block;
+  padding: var(--app-bar-padding);
+  font-weight: var(--app-bar-brand-font-weight, 600);
+  color: inherit;
+  text-decoration: none;
+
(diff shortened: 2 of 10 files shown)


viktor-yakubiv:
@Joozty after merging this PR and or without #48 I would like to publish a package.

Do you have any other requirement you need from the @oacore/design package for this moment?

src/elements/app-bar/styles.css
```diff
@@ -0,0 +1,37 @@
+:root {
+  --app-bar-padding-x: var(--component-padding-x, 1rem);
+  --app-bar-padding-y: var(--component-padding-y, 1rem);
+  --app-bar-padding: var(--app-bar-padding-y) var(--app-bar-padding-x);
+  --app-bar-background: var(--white, #fff);
+  --app-bar-color: var(--gray-800, inherit);
+  --app-bar-hover-background: #0000000a;
+}
```
Joozty: Isn't it possible to define these variables only on module scope level?

src/utils/class-names.js
```diff
@@ -21,21 +21,39 @@ class ClassNames {
...
   }
 
-  withModule(module) {
+  from(module) {
     Object.assign(this.module, module)
     return this
   }
 
+  /**
+   * @deprecated
```
Joozty: Couldn't you remove it completely?

src/elements/app-bar/brand.jsx
```diff
@@ -0,0 +1,28 @@
...
+      .from(styles)
+      .join(className)}
+    {...restProps}
+  >
+    {children}
+  </Tag>
+)
+
+Brand.propTypes = {
+  color: PropTypes.oneOf(['default', 'primary', 'success', 'danger']),
```
Joozty: color isn't used, is it?

src/elements/app-bar/brand.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import styles from './styles.css'
+
+import { classNames } from 'utils'
+
+const Brand = ({ children, className, tag: Tag = 'a', ...restProps }) => (
```
Joozty: I was expecting to have logo here too.

Joozty:
I was expecting to have [Item component](https://github.com/oacore/dashboard/blob/master/design/header/index.jsx#L12) in AppBar too. Was there any reason why you didn't include it. 

To be honest I didn't want to move app bar here until it is completely done in @oacore/dashboard but I am ok with it.

src/elements/app-bar/brand.jsx
Joozty: Ohh sorry I missed this one https://github.com/oacore/design/issues/49

src/elements/app-bar/brand.jsx
viktor-yakubiv: Thank you

src/elements/app-bar/styles.css
viktor-yakubiv: It's possible but this would be too much. I do not want to do it now. Should we do so? If yes, why?

viktor-yakubiv:
I forgot to add _Avatar_. I did not want to add _Item_ but I find it useful for this moment.

I will update PR. Thank you for the review.

src/utils/class-names.js
viktor-yakubiv: I could but as a library developer I cannot be sure that this is not used anywhere, so I have to keep it.

Also, such approach allows me to migrate internal components later.

src/elements/app-bar/styles.css
Joozty: I was just curious. It is not necessary. I just think we have a lot of global variables and it can cause troubles in the future.

viktor-yakubiv: Not really. It allows you config the framework even if you include configuration after source files. I like it more.

In other words, if you do the following:

```css
@import url('~@oacore/design/lib/index.css');

:root {
  --app-bar-background: var(--gray-100);
}
```

the AppBar will become light grey.

src/elements/app-bar/brand.jsx
viktor-yakubiv: It's not expected. You can pass whatever as a children but the logo is not this component's responsibility.

src/utils/class-names.js
Joozty: I couldn't agree more but let's be honest here. We use this library only in @oacore/dashboard and we agreed it's a beta version. And everything may change. I propose to do this stuff once we release 1.0.0.

src/elements/app-bar/styles.css
Joozty: Some of the variables are useful some not. I can imagine changing background color but I do not really care about padding you know.

viktor-yakubiv:
@Joozty is this what you wanted to have?

I decided that I am not going to add _AppBar.Avatar_ because it is not an element, i.e. pure wrapper around HTML element but more complex component that consists from:
- _AppBar.Item_ or not-existing _AppBar.Button_
- _Avatar_ that is a wrapper around ``
- hidden (screenreader only) text caption with username or button action

src/utils/class-names.js
viktor-yakubiv: I agree. On other hand, I have to modify all other component in all other PRs here to make it work. I do not want that. It worth to keep `@deprecated` for a few commits, start using new API and the migrate at some point all other components.

PR #53, opened 2019-12-13
Fix font face based on custom property

viktor-yakubiv:
Adds `--font-family-base` to fix missing font face.

Removes redundant font-family duplication from styleguidist
configuration.


viktor-yakubiv:
@Joozty this will fix font face in oacore/dashboard after deployment.

Issue #54, opened 2019-12-13
Migrate to sanitize.css


viktor-yakubiv:
Replace normalize.css to [sanitize.css](https://csstools.github.io/sanitize.css/). It's more what we want.

In case of doing that we can remove custom Bootstrap Reboot.

PR #56, opened 2019-12-14
Use sanitize.css

Joozty:
Closes https://github.com/oacore/design/issues/54


src/index.css
```diff
@@ -1,3 +1,6 @@
+@import url('~sanitize.css');
+@import url('~sanitize.css/forms.css');
```
viktor-yakubiv: I think, we need all of it, typography too.

PR #59, opened 2019-12-17
AppBar fixes

viktor-yakubiv:
A few small fixes after the release.


src/elements/index.js
```diff
@@ -1,5 +1,8 @@
+export AppBar from './app-bar'
+export * from './app-bar'
 export Heading from './heading'
 export Button from './button'
 export * from './forms'
 export Label from './label'
+export Table from './tables'
```
Joozty: Why did you change it? It works just fine for me in https://github.com/oacore/dashboard/pull/20

Joozty: You export it two times, don't you?

viktor-yakubiv: It's according to the spec. If I export *, it does not include default. So I do 2 exports, one for default and one for the rest

Joozty: Anyway it's exported here so it doesn't really matter https://github.com/oacore/design/blob/master/src/elements/tables/index.js#L16

viktor-yakubiv: Read carefully
https://github.com/oacore/design/pull/59/files#diff-a0a697338eb47b168a0854933a67eaa4L16

Joozty: Sorry, my bad 😴

PR #84, opened 2020-02-26
Add new icons from @oacore/dashboard

Joozty:


viktor-yakubiv:
I see no need in this at the moment but feel free to merge if you need it it

Joozty:
I need it because I want to use settings icon and this icon is in different sprite then all other menu icons. And I don't want to add a custom condition to activity select.

PR #124, opened 2020-03-20
Fix classnames

Joozty:


src/utils/class-names.js
```diff
@@ -27,7 +27,8 @@ class ClassNames {
   }
 
   join(className) {
-    this.appendix = className
+    if (className instanceof ClassNames) this.appendix = className.toString()
+    else this.appendix = className
```
viktor-yakubiv: Do not want you to explicitly convert this and the above to string? This may deprecate `join()` at all

```js
  use(...classList) {
    this.classList.push(...classList.map(item => item.toString()))
    return this
  }
```

Joozty: I think you created join because it could happen that the module contained the same class name as it was passed down. (maybe some global class name or whatever) Am I wrong? I mean it is an edge case but still, we should keep it in mind.

viktor-yakubiv: Yes, this is correct. Sorry

PR #158, opened 2020-04-08
Add helper message to TextFields

Joozty:


src/elements/forms/text-field.css
```diff
@@ -64,12 +78,53 @@ label {
...
+  }
+}
+
+input:invalid,
+input:invalid + label {
+  --form-control-color: var(--danger-dark);
+
+  /* Form some reason --form-control-color change is not enough */
+  --form-label-color: var(--form-control-color);
+}
```
viktor-yakubiv, line 102: I do not think we should have this as a default behaviour. CSS validates during the input, JS validates on blur or submit. It's inconsistent and we should choose.

Since we develop a library, choosing ourselves would be opinionated what is the wrong approach for a library that pretends to be flexible (more or less) unless it's Prettier 😃

src/elements/forms/text-field.css
```diff
@@ -32,44 +42,93 @@ input {
...
+  &:empty {
+    display: none;
+  }
+}
+
+input:invalid,
+input:invalid + label {
+  --form-control-color: var(--danger-dark);
+
+  /* Form some reason --form-control-color change is not enough */
```
Joozty, line 84:
```suggestion
  /* For some reason --form-control-color change is not enough */
```

viktor-yakubiv:
Please pay some attention to commits. You asked me to finish the PR. I did it. Then you review, add a small change and merge with squash.

I did 2 separate commits for a reason. One was in the PR's scope, another was out of but slightly touched the scope.

The line you edited was repeated **three** times but you committed only single one.

PR #160, opened 2020-04-13
Add option icons for the new Card design

viktor-yakubiv:
Adds:

- `alert-circle` and `check-circle` (for highlighting statuses)
- `close` (for dialogs)
- `cog` (instead `settings` that is deprecated now)
- `dots-vertical` (menu button)
- `help-circle` and `help-circle-outline` (contextual help)
- `magnify` (search)
- `tune` (preferences)

Icons are copied from `@mdi/svg` package.

---

@Joozty these are not optimised for a reason. I want to remove icons from the repository and set sprite building up from the [@mdi/svg](https://github.com/Templarian/MaterialDesign-SVG).


Joozty:
@viktor-yakubiv Why don't you just do it instead of copying these icons here and then removing it later. Is it hard to do it?

viktor-yakubiv:
Because we have to deliver things fast. The described solution is a breaking change and will require changes in other projects accordingly. It will stop other projects for a moment because the update is a need at the moment.

PR #166, opened 2020-04-14
Build icons using config

viktor-yakubiv:
I am trying to write a script that would do exactly the same what we have in the `package.json`:

```
svgo -f src -o lib -r --disable=removeViewBox --enable=removeDimensions",
svgstore lib/assets/icons/* | svgo -i - -o lib/assets/icons.svg --disable=removeUselessDefs,cleanupIDs
```

SVGO is a quite tricky thing to manage it running from the node script ~. I am struggling to do optimisations. Apart from that~, everything works.

### Roadmap

- [x] Create icons build script
- [x] Read options from the config
- [x] Add cosmiconfig
- [ ] Add yargs or coa


bin/icons/index.js
```diff
@@ -0,0 +1,82 @@
...
+    removeUselessDefs: false,
+    cleanupIDs: false,
+  },
+}
+
+const run = async () => {
+  const config = await readConfig()
+  const paths = await resolveInput(config.input)
+  const files = await read(paths)
+  const icons = await optimize(files, sourceOptions)
```
viktor-yakubiv, line 77: For some reason, SVGO removes all content leaving only `<svg>` tag. I have no idea what the problem may happen.

config.js
```diff
@@ -0,0 +1,65 @@
...
+  'check-circle',
+  // 'cite',
+  'close',
+  'cog',
+  // 'core-symbol',
+  // 'dashboard',
+  // 'data',
+  'file-alert', // 'document-alert',
+  'file-check', // 'document-success',
+  'file', // 'document',
```
viktor-yakubiv, line 16: @Joozty be aware this is a breaking change. `core-symbol` should not be an icon. The rest will be removed or renamed according to the original icon names from `@mdi/svg` what you may see in the current file. I am working now on finding an proper replacement for everything.

In the future (maybe this, maybe next PR unless we discuss it) design will not ship any icon. The dependent project should build own sprite using provided tools.

viktor-yakubiv:
My working directory after some testing 😄

a lot of svg directorie

config.js
viktor-yakubiv: I added full mapping.

viktor-yakubiv:
I will not be adding `yargs` or `coa`. It's not necessary at the moment and current result suffices me. @Joozty could you please review?

I will remove current icons after and remove the default config in a separate PR.

bin/icons/config.js
```diff
@@ -0,0 +1,101 @@
...
+
+const read = async (filepath) => {
+  const explorer = cosmiconfig('design')
+  if (filepath != null) return explorer.load(filepath)
+
+  let config = await explorer.search()
+  if (config == null) config = await explorer.load(defaultFilepath)
+
+  return config
+}
```
Joozty, line 39:
```suggestion
const read = async (filepath = defaultFilepath) => {
  const explorer = cosmiconfig('design')
  return explorer.load(filepath)
}
```

bin/icons/config.js
```diff
@@ -0,0 +1,101 @@
...
+  const output = prepareOutput(config.output)
+  const svgo = { ...defaultConfig.svgo, ...config.svgo }
+  return { icons, output, svgo }
+}
+
+const load = async (filepath) => {
+  const { config: userConfig } = await read(filepath)
+  const config = prepare(userConfig)
+  return config
+}
```
Joozty, line 99:
```suggestion
const load = async (filepath) => {
  const { config: userConfig } = await read(filepath)
  return prepare(userConfig)
}
```

viktor-yakubiv: I wanted to leave this as it is. It allows putting `console.log` in between without changing anything.

bin/icons/config.js
viktor-yakubiv: It's not like that. I want to load the default `design` config only if there is not project-specific one.

viktor-yakubiv:
Thank you! I used a lot of type checks to make the configuration as flexible as possible.

PR #170, opened 2020-04-24
Create global design command with Yargs

viktor-yakubiv:
As simple as:
```sh
npx @oacore/design build-icons
```

I am so excited about 🎉


viktor-yakubiv:
I do not mind how others run the binary. However, from this point of view, I should remove `@mdi/svg` from dependencies when I stop shipping icons with the design.

PR #171, opened 2020-04-24
Build runtime config

viktor-yakubiv:
This adds a public path resolver for icons. The public path to the sprite is stored under the `lib/config.js` and can be imported to any component.

Ideally, it should be inlined to keep components independent from central files but we can start from this solution.

Find out more in the commit messages.


build/config/index.js
```diff
@@ -0,0 +1,52 @@
...
+const genrateValue = (stringOrNull) =>
+  stringOrNull == null ? 'null' : `"${escapeString(stringOrNull)}"`
+
+// generated with Babel, edited by hand
+// variables substituted manually
+const generatePublicConfig = ({ publicPath, iconsPublicPath }) => `
+"use strict";
+
+/**
+ * This file is automarically generated by the binary command running
```
Joozty:
```suggestion
 * This file is automatically generated by the binary command running
```

viktor-yakubiv:
The usage is as simple as:
```sh
npx @oacore/design build icons
npx @oacore/design build config

# shortcut for both commands
npx @oacore/design build
```

Also, it allows to call each file as a command manually:

```sh
# the following lines are equal to 'npx @oacore/design build icons'
node node node_modules/bin build icons
node node_modules/bin/build icons
node node_modules/bin/build/icons
```

PR #173, opened 2020-04-27
Add icon sprite resolver

viktor-yakubiv:
Resolves icon sprite references from iconsPublicPath if the the source path starts from the '#'.


bin/build/icons/runner.js
```diff
@@ -99,6 +104,7 @@ const run = async () => {
     const dirname = path.dirname(outputConfig.sprite)
     console.log(`${basename} written to ${dirname}`)
   }
+  console.log('')
```
Joozty, line 107: Is this intended?

PR #175, opened 2020-04-30
Move placeholder up when input is autofilled

Joozty:
For some browsers `:not(:placeholder-shown)` is not triggered hence this new rule


Joozty:
@viktor-yakubiv I put one change after you reviewed. I hope it's ok.

viktor-yakubiv:
It was not okay. I had a few questions with further changes

PR #179, opened 2020-05-11
Allow to hide label for TextFields

Joozty:


src/elements/forms/text-field.jsx
```diff
@@ -29,6 +29,7 @@ const TextField = React.forwardRef(
       helper,
       variant = 'normal',
       size = 'default',
+      labelSrOnly = false,
```
viktor-yakubiv, line 32: `labelless` maybe?

Issue #178, opened 2020-05-07
Style readonly input fields


Joozty:
Currently, the read-only input fields look the same as writable. Would be nice to distinguish them and maybe grey them out?

PR #180, opened 2020-05-11
Style disabled TextFields

Joozty:
Closes https://github.com/oacore/design/issues/178

### DIFF
--- a/src/elements/forms/text-field.css
+++ b/src/elements/forms/text-field.css
@@ -77,6 +77,16 @@ label {
   }
 }
 
+input:disabled,
+input:read-only,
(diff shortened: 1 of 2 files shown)


viktor-yakubiv:
It's a good start. However, the read-only should not be 'disallowed'

Joozty:
Why? I see both of them the same.

viktor-yakubiv:
Read-only allows to select and copy text but does not allow to change. `Disabled` means that the user must not interact with. It should be clear.

Joozty:
Interesting. Thanks for the explanation.

PR #186, opened 2020-06-09
Add missing monospace font

viktor-yakubiv:


viktor-yakubiv:
@Joozty it's my fault it was not there. I am not sure it will not break anything if we add this now. Maybe this should go to the next release. What do you think?

Joozty:
Where is this variable used?

viktor-yakubiv:
`variant="pure"` removes the label to a sr-only.

I was thinking about having a specific behaviour when the label fades out on focus but remains the rest of the time on the front of the input that is why I introduced a variant. Another reason is that I do not really like `labelSrOnly` property.

viktor-yakubiv:
The variable is used for `code`, `kbd`, `pre` and other monospace elements.

PR #193, opened 2020-06-15
Improve TextField API and styles

viktor-yakubiv:
Introduces class-first approach for the inputs.

Fixes the bug with the Firefox broken label animation by removing the `-webkit-autofill` selector.

---

@Joozty I wanted to fix the bug and ended up rewriting everything. Even if it seems not worth it, it is for me. This may be the first step of writing the next forms API in the library. Let me know what do you think about.

### DIFF
--- a/src/elements/forms/text-field.css
+++ b/src/elements/forms/text-field.css
@@ -2,8 +2,10 @@
   /* Control */
   --form-control-corner-radius: var(--component-corner-radius, 2px);
   --form-control-border-width: 1px;
+  --form-control-font-size: var(--body-font-size, 1rem);
   --form-control-padding-x: var(--component-padding, 1rem);
   --form-control-padding-y: var(--component-padding, 1rem);
+  --form-control-background-color: var(--white);
   --form-control-color: var(--primary);
 
   /* Label */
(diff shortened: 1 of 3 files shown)


src/elements/forms/text-field.css
```diff
@@ -77,52 +137,49 @@ label {
...
-  background: var(--white, #fff);
-  transform: translateY(
-    calc(
-      -1 * var(--form-control-padding-y) - var(--form-label-font-size) / 2 - var(--form-control-border-width)
-    )
-  );
+.pure {
+  /* Kepping screenreader-only tweak as is (with !important-s) for a safety */
+  /* stylelint-disable declaration-no-important */
+  & > .label {
```
Joozty, line 154: This is duplicate code. What about applying sr-only class on JS level? :)

viktor-yakubiv:
I fixed Stylint issues and rebased to solve conflicts. I hope I did not break anything and wonder what changes could happen with the text field today.

Joozty:
@viktor-yakubiv just some empty lines were removed from there.

Joozty:
As we already talk about it the autofill is crucial here and it should stay in there because otherwise chrome is broken. Apart from that refactor is wonderful. :)
pass

viktor-yakubiv:
@Joozty I tried to return autofill but the only thing I know that it does not break Firefox now. Could you please check it?

I decided to duplicate the code for now and use the mixins later when we add tooling for it.

viktor-yakubiv:
@Joozty I am only not sure about having success/error for a 'pure' TextField. Currently, it's not possible but it should be possible probably. We have to distinguish layout and colour variants but I am struggling with consistency vs new API decisions.

Issue #197, opened 2020-06-17
Develop more flexible form controls


viktor-yakubiv:
@Joozty I would like to discuss the following interface:

```jsx
<Form.Group>
  <Form.Label>User name</Form.Label>
  <Form.Control value="john" placeholer="e.g. johndoe" />
  <Form.Control.PrependIcon src="user" />
  <Form.Text>Nice username</Form.Text>
</Form.Group>
```

It's a wrapper around HTML. However, it applies classes and can possibly sort elements, i.e. put the label in the end, after the input.

The TextField component will remain but would be a shortcut for this expanded group.

Joozty:
I am quite neutral here. I understand we need more control for inputs, on the contrary, we would need 6 lines of code to develop simple input. I like TextFields for their simplicity. So inclining more to use this only internally and the user should use it only in edge cases.

All the props and children in your example are just strings. I doubt it can be anything else. So it's the same as using just new props for TextField.


```
<TextField
 label="User name"
 value="john"
 placeholder="e.g. johndoe"
 icon="#success"
 text="Nice username"
/>
```

PS: My example has 7 lines of code. It's even worse 🤦 It was not a good idea to measure code lines. 😄

viktor-yakubiv:
I still would prefer your example if I use the library somewhere. I have only one remark to your code: there may be more than a single icon and the interface is not clear enough yet.

I see 2 options:
- `icon` is the left any icon, `status: ['check', 'alert', 'loading']` is the right icon with the limited options
- `prependIcon` for the left one, `appendIcon` for the right one, both can be any

However, this issue is only about the internal flexible interface to easily adjust our styles to selects and maybe other fields and provide some interface for edge cases as you said.

Do you have any suggestions to the interface itself? I do not like this bit:

```jsx
 <Form.Control.PrependIcon src="user" />
```

Joozty:
Do we really want to have two icons? If so then your first opinion looks better for me.

I think you picked a reasonable naming. I would maybe just call Form.Text -> Form.Description and rename Form -> Input but that's like really opinionated and minor...

viktor-yakubiv:
I will start developing it like it was designed originally and then we will see what naming would be the best. Thank you for the feedback 👍

viktor-yakubiv:
@Joozty do you have any libraries for the inspiration you like? I checked [ant.design](http://ant.design), I do not really like the API but it's not bad.

[...]

PR #198, opened 2020-06-17
Develop form control

viktor-yakubiv:
Adjusts TextField styles to be more flexible for the Select
and TextArea elements:
* adds line-height
* reduces vertical padding
* includes new variables to the calculations

Potential problems:
* `line-height` changes with the label during the sliding animation, what may lead to performance issues

Adds `FormControl` element as a wrapper around simple `input`, `textarea` and `select`. @Joozty do I miss any other element should be covered?

### DIFF
--- a/src/elements/forms/styles.css
+++ b/src/elements/forms/styles.css
@@ -1,0 +1,26 @@
+:root {
+  --textarea-min-line-count: 1;
+  --textarea-line-count: 3;
+  --form-control-line-height: var(--line-height-base, 1.5);
+}
+
+/* Restore height for the text area */
+textarea {
+  min-height: calc(
+    2 * var(--form-control-padding-y) + var(--form-control-line-height) *
+      var(--textarea-min-line-count) * var(--form-control-font-size)
+  );
+  height: calc(
+    2 * var(--form-control-padding-y) + var(--form-control-line-height) *
+      var(--textarea-line-count) * var(--form-control-font-size)
+  );
+  padding-top: calc(var(--form-control-padding-y) - 1px);
+  padding-bottom: calc(var(--form-control-padding-y) - 1px);
+  line-height: var(--line-height-base, 1.5);
+  height: auto;
+}
+
+select:not([multiple]):not([size]) {
+  background-origin: content-box;
+  background-position: right center;
+}
--- a/src/elements/forms/styles.css
+++ b/src/elements/forms/styles.css
@@ -6,21 +6,21 @@
 
 /* Restore height for the text area */
 textarea {
-  min-height: calc(
-    2 * var(--form-control-padding-y) + var(--form-control-line-height) *
-      var(--textarea-min-line-count) * var(--form-control-font-size)
-  );
   height: calc(
     2 * var(--form-control-padding-y) + var(--form-control-line-height) *
       var(--textarea-line-count) * var(--form-control-font-size)
   );
+  height: auto;
+  min-height: calc(
+    2 * var(--form-control-padding-y) + var(--form-control-line-height) *
+      var(--textarea-min-line-count) * var(--form-control-font-size)
+  );
   padding-top: calc(var(--form-control-padding-y) - 1px);
   padding-bottom: calc(var(--form-control-padding-y) - 1px);
   line-height: var(--line-height-base, 1.5);
-  height: auto;
 }
 
 select:not([multiple]):not([size]) {
-  background-origin: content-box;
   background-position: right center;
+  background-origin: content-box;
 }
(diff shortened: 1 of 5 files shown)


src/elements/forms/styles.css
```diff
@@ -0,0 +1,26 @@
...
+  );
+  padding-top: calc(var(--form-control-padding-y) - 1px);
+  padding-bottom: calc(var(--form-control-padding-y) - 1px);
+  line-height: var(--line-height-base, 1.5);
+}
+
+select:not([multiple]):not([size]) {
+  background-position: right center;
+  background-origin: content-box;
+}
```
viktor-yakubiv, line 26: This is the tricky part. Select is styled by [applying a triangle icon on the background](https://github.com/csstools/sanitize.css/blob/5bd695fdc7be25221c542810a72c174a33d6fb00/forms.css#L35). The current code shifts this to the content box to skip 1rem (16 pixels) from the right.

The problems are:
1. we should apply our own icon from `@mdi/svg`
2. the icon should be styled and ideally should change the colour with the border colour
3. ideally it should be applied with CSS too what makes it even more tricky
    - the trickiness is that if we apply the icon with CSS, it should be so everywhere
    - we probably should support different colour icons but maybe not, I am not sure yet
    - if the icon is applied with CSS, JS API will be different from proposed in #197 originally
4. the icon and the content should not overlap, so shifting to the content is not correct
    - the padding-right should be expanded in 3 times: `16px` for padding, `16px` for the icon box, `16px` more for the padding between the icon and the actual text box
    - in this case, positioning is not so easy since background aligns [very specifically](https://developer.mozilla.org/en-US/docs/Web/CSS/background-position)

src/elements/forms/text-field.css
```diff
@@ -3,8 +3,9 @@
   --form-control-corner-radius: var(--component-corner-radius, 2px);
   --form-control-border-width: 1px;
   --form-control-font-size: var(--body-font-size, 1rem);
+  --form-control-line-height: var(--line-height-base, 1.5);
```
viktor-yakubiv, line 6: It seems like this introduces more issues than solves. It makes text-area size calculation more implicit and easy but at the same time, it (maybe) affects the performance of all the inputs adding another variable into calculations.

src/elements/forms/styles.css
Joozty: > we probably should support different colour icons but maybe not, I am not sure yet

IMHO icon should always be dependent on `currentColor`.


> in this case, positioning is not so easy since background aligns very specifically
Why not to have it as `:before` and `:after` element. Anyway if possible it would be better if icons are not in CSS so a user can use any icon they want.

I would say there is one more problem - **icons should be part of sprite ideally**.


It's a tough task. Thanks for working on it. 🍀 🤞

Joozty:
> What the worst but real usage of the component (with icons specifically) do you see?

I don't know. I would input with custom icon on the right and custom icon on the left. Btw. I would suggest to not support custom right icon. This icon should be picked automatically based on whether field is valid or not.

src/elements/forms/text-field.css
Joozty: I wouldn't be worried about performance. 🙂

src/elements/forms/styles.css
viktor-yakubiv: > icon should always be dependent on `currentColor`.

It's not like that. Here it should be `var(--form-control-color)` specifically. The `currentColor` is the value of `color` property in the current element, i.e. it's the text colour, i.e. black, what does not help us.

I could use the background blend mode if it would work for icons but I am afraid not. I will try anyway.

> icons should be part of sprite ideally

Ideally yes. Technically it does not work. CSS is tied to HTML, you can only reference SVG symbols written in the HTML but cannot use external ones.

It's possible to programmatically build a CSS file with all icons in it exposed as classes or variables. It might work but I am not completely sure in the need of this.

> Why not to have it as `:before` and `:after` element.

These pseudo-elements are not available for the `<input>` element. I would do so if it would be possible 🙂

The decision should rely on the way of using the icons. I see a presentational role for it only. It does not require any accessibility because the text label always exist and for valid/invalid statuses we have special reporting too. Icons are complementary only and can be part of the background.

viktor-yakubiv:
I see that there may be a problem when we want to have one icon repeating the form colour (the left side, eg. search or telephone icon) and another one having a different colour which should not change if the input focus status change, e.g. loading (spinner) icon should be always grey. I will see, what's possible to do.

viktor-yakubiv:
Oops, sorry. I wanted to merge this to another branch

PR #199, opened 2020-06-22
Develop more flexible form components

viktor-yakubiv:
This should include:
* [x] FormControl (#198)
* [x] FormLabel (#201)
* [x] Form
* [ ] FormText

and some other components for icons, statuses

### DIFF
--- a/design.config.js
+++ b/design.config.js
@@ -28,12 +28,14 @@ const path = require('path')
 const isDev = process.env.NODE_ENV !== 'production'
 
 const icons = [
+  'alert-circle-outline',
+  'check',
   'open-in-new',
   path.resolve(__dirname, './assets/core-symbol.svg'),
 ]
 
 // icons used in documentation
-if (isDev) icons.push('check', 'file-check')
+if (isDev) icons.push('file-check')
 
 const iconsRoot = path.join(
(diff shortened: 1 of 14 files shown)


viktor-yakubiv:
We also need an `InputAddon` component. I think that `FormControl` should be renamed to `Input` later.

- [ ] InputAddon
- [ ] Rename `FormControl` to `Input`

viktor-yakubiv:
I don't think we need a `Form` component, maybe for the consistency only.

I will probably rename the `FormControl` to the `Input` for the consistency with the `InputAddon`.

I am not happy about the API but it seems we keep using TextField only and that's all. However, should continue working on this because we need much more for the Search:
- Menu for the autocomplete
- Select (custom and very generic one) for the Dashboard and maybe for the Seach too
- Checkbox for various settings
- Radiobutton 🤔 I am not sure
- Switch for settings too

viktor-yakubiv:
I fixed conflicts and going to add a few edits to make this finished

viktor-yakubiv:
@Joozty I hope this is the final stage. Feel free to review only last 3 commits since you already checked the rest in various PRs.

Since I resolved conflicts in the `design.config.js` you can check the file for everything being done right.

PR #202, opened 2020-06-23
Add the FormGroup component

viktor-yakubiv:
It'a simple wrapper around the `<p>` element that adds another class. It will be richer in the future.

### DIFF
--- a/src/elements/forms/styles.css
+++ b/src/elements/forms/styles.css
@@ -20,6 +20,8 @@
   --textarea-line-count: 3;
 }
 
+/* Generics */
+
 %label {
   display: block;
   margin-bottom: 0.5rem;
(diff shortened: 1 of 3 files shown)


src/elements/forms/styles.css
```diff
@@ -108,10 +110,63 @@
...
+ * FormLabel
+ * =============================================================================
+ */
+
+.label {
+  @extend %label;
+}
+
+/**
+ * FormControl
```
Joozty, line 123: Maybe you should use regions comments. It allows the collapsing of the whole region at once. Not sure if it is supported in your editor I hope so. :) 

/* region: FormControl */
code here
/* endregion */

viktor-yakubiv: I am not a fan of such comments. If you do not mind, I will not add it.

Personally, I prefer having it in the separate files but we need to explore CSS Modules config more deeply to make hashes be the same for all files within the directory. I know the way to do this but I do not think it's a need at the moment.

PR #204, opened 2020-06-23
Add the ProgressSpinner component

viktor-yakubiv:
The component can display indeterminate progress or indicate a specified value.

Added examples of the 2 use cases.

---

The component is the first step to progress indicating for the inputs.


src/elements/progress-spinner/progress-spinner.md
```diff
@@ -0,0 +1,51 @@
...
+        <ProgressSpinner
+          value={value}
+          max={maxValue}
+          style={{ marginRight: '0.5rem' }}
+        />
+        Loading...
+      </p>
+      <Button variant="contained" onClick={() => { setValue(0) }} disabled={value < maxValue}>
+        Restart
+      </Button>
```
Joozty, line 45: Is it intentional then loading animation starts from the quarter?

src/elements/progress-spinner/progress-spinner.md
```diff
@@ -0,0 +1,51 @@
...
+### Controlled
+
+Indicates the actual progress value.
+
+```jsx
+import { useEffect, useRef, useState } from 'react'
+import Button from '../button';
+
+const AnimatedProgress = () => {
+  const timeoutRef = useRef()
```
Joozty, line 18:
```suggestion
  const timeoutRef = useRef(null)
```

src/elements/progress-spinner/progress-spinner.md
```diff
@@ -0,0 +1,51 @@
...
+  
+  useEffect(() => {
+    if (value < maxValue) {
+      timeoutRef.current = setTimeout(() => {
+        setValue(value + 1)
+      }, 20)
+    }
+    
+    return () => clearTimeout(timeoutRef.current)
+  })
```
Joozty, line 31:
```suggestion
  }, [value])
```

src/elements/progress-spinner/progress-spinner.md
viktor-yakubiv: It's a trick I do not really know how to deal with in the example. When you click 'restart', the value immediately changes to 0 and starts increasing by 1 every 20 milliseconds. In the same time, we have a transition for 250 milliseconds that starts immediately after the button click too.

The `transition` decreases the actual visual value by 1 every `2.5ms` when the `setTimeout` increases the value by 1 every `20ms`. Putting it together: `2.5 × (100 - x) = 20x` → They meet at the `x = 11` (11% of the path from the start).

src/elements/progress-spinner/progress-spinner.md
viktor-yakubiv: I am not sure this is what I want. Anyway, the component does not render between changing values and the effect triggers value change anyway.

How do you see this? Why adding the value dep would be more correct?

src/elements/progress-spinner/progress-spinner.md
viktor-yakubiv: I do not really see a big difference between `null` and `undefined` in this case

src/elements/progress-spinner/progress-spinner.md
viktor-yakubiv: Although, it does not work as expected and it is actually after the quarter. I have no idea why

src/elements/progress-spinner/progress-spinner.md
Joozty: There is no difference you are right. I just prefer explicitly put param to that hook because if you have some type check it shows you a warning.

src/elements/progress-spinner/progress-spinner.md
Joozty: The same as [here](https://github.com/oacore/design/pull/204#discussion_r444811850). It's just correctness of hook usage nothing else. If you extend your example (adding new state) in the future it won't work as expected. :)

src/elements/progress-spinner/progress-spinner.md
Joozty: Me neither to be honest. This SVG stuff is very new to me. 😕

PR #206, opened 2020-07-06
Allow to use own design config in the repo

Joozty:
This PR adds support for default config so we can use custom icons here too.

The flow is the following:
1. Load config from this repo and resolve icons with absolute path
2. Search for config in the project root `(/<project_root>/node_modules/@oacore/design/bin)`
3. Resolve icons in this config
4. Merge configs together

### DIFF
--- a/bin/config.js
+++ b/bin/config.js
@@ -1,5 +1,6 @@
 const path = require('path')
 
+const merge = require('deepmerge')
 const { cosmiconfig } = require('cosmiconfig')
 
 const defaultConfig = {
@@ -27,15 +28,6 @@ const doc = {
     'config.icons should be either string, array or an object { files, path }',
 }
 
-const read = async (filepath) => {
-  const explorer = cosmiconfig('design')
-  if (filepath != null) return explorer.load(filepath)
-
-  const config = await explorer.search()
-
-  return config == null ? { config: defaultConfig } : config
-}
-
 const appendExtension = (file, ext) =>
   new RegExp(`${ext}$`).test(file) ? file : `${file}${ext}`
 
@@ -54,49 +46,59 @@ const prepareIcons = (input) => {
   return files
 }
 
-const prepareOutput = (output) => {
-  const result = {
-    ...defaultConfig.output,
-    icons: { ...defaultConfig.output.icons },
-  }
+const mergeAndPrepareConfigs = (...configs) => {
+  // Create absolute path for icons so they can be merged
+  configs.forEach(({ config }) => {
+    if ((config || {}).icons) config.icons = prepareIcons(config.icons)
+  })
+  const mergedConfig = merge.all(configs)
+  const {
+    config: { output },
+  } = mergedConfig
+  if (output.icons.files)
+    output.icons.files = path.resolve(output.path, output.icons.files)
 
-  if (typeof output == 'string') result.path = output
-  else {
-    if (output.path) result.path = output.path
-    if (output.publicPath) result.publicPath = output.publicPath
+  if (output.icons.sprite)
+    output.icons.sprite = path.resolve(output.path, output.icons.sprite)
 
-    if (typeof output.icons == 'boolean')
-      result.icons = { ...defaultConfig.output.icons }
-    else if (typeof output.icons == 'string') {
-      result.icons = {
-        files: /.svg$/.test(output.icons) ? null : output.icons,
-        sprite: appendExtension(output.icons, '.svg'),
-      }
-    } else result.icons = { ...output.icons }
-
-    if (result.icons.files)
-      result.icons.files = path.resolve(result.path, result.icons.files)
-    if (result.icons.sprite)
-      result.icons.sprite = path.resolve(result.path, result.icons.sprite)
-  }
-
-  return result
+  return mergedConfig
 }
 
-const prepare = (config) => {
-  const icons = prepareIcons(config.icons)
-  const output = prepareOutput(config.output)
-  const svgo = { ...defaultConfig.svgo, ...config.svgo }
-  return { icons, output, svgo }
+const loadConfigs = async () => {
+  const explorer = cosmiconfig('design', {
+    stopDir: '/',
+  })
+
+  // load all configs
+  const configs = []
+  let searchFrom = process.cwd()
+  do {
+    // eslint-disable-next-line no-await-in-loop
+    const result = await explorer.search(searchFrom)
+    if (result) {
+      configs.push(result)
+      // to avoid infinite loop when config is places under /design.config.js
+      if (searchFrom === '/') break
+      searchFrom = path.resolve(path.dirname(result.filepath), '../')
+    } else searchFrom = null
+  } while (searchFrom != null)
+
+  return configs
 }
 
-const load = async (initialFilepath) => {
-  const configResult = await read(initialFilepath)
-  const { filepath: actualFilepath, config: userConfig } = configResult
-  const config = prepare(userConfig)
+const load = async () => {
+  const cosmicConfigs = await loadConfigs()
+  const { config } = mergeAndPrepareConfigs(
+    { config: defaultConfig },
+    ...cosmicConfigs
+  )
 
-  if (actualFilepath) console.log(`Using config from ${actualFilepath}`)
-  else console.log('Using default config')
+  if (cosmicConfigs.length) {
+    console.log(`Merged ${cosmicConfigs.length} configs from:`)
+    cosmicConfigs.forEach(({ filepath }) => {
+      console.log(`\t ${filepath}`)
+    })
+  } else console.log('Using default config')
 
   return config
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -5896,10 +5896,9 @@
       "dev": true
     },
     "deepmerge": {
-      "version": "1.3.2",
-      "resolved": "https://registry.npmjs.org/deepmerge/-/deepmerge-1.3.2.tgz",
-      "integrity": "sha1-FmNpFinU2/42T6EqKk8KqGqjoFA=",
-      "dev": true
+      "version": "4.2.2",
+      "resolved": "https://registry.npmjs.org/deepmerge/-/deepmerge-4.2.2.tgz",
+      "integrity": "sha512-FJ3UgI4gIl+PHZm53knsuSFpE+nESMr7M4v9QcgB7S63Kj/6WqMiFQJpBBYz1Pt+66bZpP3Q7Lye0Oo9MPKEdg=="
     },
     "default-gateway": {
       "version": "4.2.0",
@@ -16852,6 +16851,14 @@
         "deepmerge": "1.3.2",
         "mitt": "1.1.2",
         "svg-baker": "^1.7.0"
+      },
+      "dependencies": {
+        "deepmerge": {
+          "version": "1.3.2",
+          "resolved": "https://registry.npmjs.org/deepmerge/-/deepmerge-1.3.2.tgz",
+          "integrity": "sha1-FmNpFinU2/42T6EqKk8KqGqjoFA=",
+          "dev": true
+        }
       }
     },
     "svg-sprite-loader": {
@@ -16869,6 +16876,14 @@
         "svg-baker": "^1.5.0",
         "svg-baker-runtime": "^1.4.7",
         "url-slug": "2.0.0"
+      },
+      "dependencies": {
+        "deepmerge": {
+          "version": "1.3.2",
+          "resolved": "https://registry.npmjs.org/deepmerge/-/deepmerge-1.3.2.tgz",
+          "integrity": "sha1-FmNpFinU2/42T6EqKk8KqGqjoFA=",
+          "dev": true
+        }
       }
     },
     "svg-tags": {
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "@mdi/svg": "^5.3.45",
     "classnames": "^2.2.6",
     "cosmiconfig": "^6.0.0",
+    "deepmerge": "^4.2.2",
     "prop-types": "^15.7.2",
     "sanitize.css": "^11.0.1",
     "svgo": "^1.3.2",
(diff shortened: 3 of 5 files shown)


bin/config.js
```diff
@@ -36,46 +28,41 @@ const prepareIcons = (input) => {
...
-  const icons = prepareIcons(config.icons)
-  const output = prepareOutput(config.output)
-  const svgo = { ...defaultConfig.svgo, ...config.svgo }
-  return { icons, output, svgo }
+const read = async (filepath) => {
+  const explorer = cosmiconfig('design')
+  // If filepath is not supplied we start search for config from project root,
+  // The reason is to avoid loading the default config.
+  // (/<project_root>/node_modules/@oacore/design/bin)
+  const mainProjectDirectory = path.resolve(__dirname, '../../../../')
```
viktor-yakubiv, line 55: Is there no better option? Maybe in `process`, e.g. `cwd` or something.

Also, please check how cosmiconfig works. I think it starts to lookup from CWD.

bin/config.js
```diff
@@ -36,46 +28,41 @@ const prepareIcons = (input) => {
...
-      result.icons.files = path.resolve(result.path, result.icons.files)
-    if (result.icons.sprite)
-      result.icons.sprite = path.resolve(result.path, result.icons.sprite)
-  }
-
-  return result
+const mergeAndPrepareConfigs = (...configs) => {
+  // Create absolute path for icons so they can be merged
+  configs.forEach(({ config }) => {
+    if ((config || {}).icons) config.icons = prepareIcons(config.icons)
```
viktor-yakubiv, line 34: I cannot find this function.

viktor-yakubiv:
I will merge first 2 commits to simplify review. Don't you mind, @Joozty?

bin/config.js
Joozty: CWD is this path `/<project_root>/node_modules/@oacore/design/` so it's actually the same (one `../` less)

bin/config.js
Joozty: You mean `prepareIcons`?

viktor-yakubiv: Yes

Joozty: it's 15 lines up

Joozty:
If those four lines of changes simplify your review go for it.  

> I do not agree with treating the default config as a library config.

Ok

> I allowed having some shortcuts in the configuration files which are always resolved by the preparing function. If you merge your internal config with the library config, you disallow yourself having all the features of the config files. It should be duplicated here for a reason.

We didn't use them so why to have it? less code == more readable. 

> Also, I do not very like implementation in some places. Why did not you reuse existing `prepare function'. In my opinion, having a reduce with a custom merging strategy would be better than a deep-merge. Also, it would drop some extra properties we do not need.

Ok

bin/config.js
viktor-yakubiv: Sorry, I thought it's implemented by yourself but it was already there.

bin/config.js
viktor-yakubiv: I checked StackOveflow, [they recommend `require.main.filename`](https://stackoverflow.com/a/18721515/7205866). I would try it but I am not sure it would work when running via Webpack, Next and even NPM.

viktor-yakubiv: It may be safe enough but it also assumes that you do not run the command from itself what is not true. Also, we did not ship any images before but we should now. This should be solved somehow, or it is solved already... 🤔 Maybe we need some ignore rules, I am not sure.

viktor-yakubiv:
> We didn't use them so why to have it? less code == more readable.

It's not always like that, e.g. `?:` is less readable than `if/else` but I understand your point. You can simplify the API if you want to. On the other side, I wanted to provide an interface as simple as possible when you do not need to configure all the paths and go in deep. It requires documentation but I find it worth from the UX point of view.

bin/config.js
viktor-yakubiv: You also may try some magic like `require.resolve('package.json')` or something similar. ESLint does this, so exploring their code it another option.

However, I still hope, that `cosmiconfig` resolves the config from the project root, from the correct one. Also, it should give you a list when you 'explore'.

Joozty: It doesn't it starts from /<project_root>/node_modules/@oacore/design/ and goes one level up till it find config or it ends in home directory. So when I added config to this repo it always found this one and stopped the search.

So I would say we shouldn't use search at all. We should always try to find always only one config in the project root. And you are right about running the command from itself. I forgot to cover this case. I will have a look at the answer you provided. I think require.main.filename will be safe to use in this scenario.

viktor-yakubiv: > So when I added config to this repo it always found this one and stopped the search.

Yes, but it accepts `searchFrom`. So we can find the first config, and recursively do `explorer.search(path.resolve('../', curerntConfig.filepath))` or similar. I still think it's better to keep `explorer.search()` since it does a lot of magic for us.

I was afraid that if we host the package differently, i.e. without `@oacore` prefix we will need to tweak the path.

When you link the package, the path it even more different, it's not under `project_root/node_modules/` but (in my case) along with the project as a sibling. It's another pain.

Joozty: I assume it can be just change like this and the problem is solved. It will work like before. We just skip the current config. I don't think it makes sense to support multiple configs. I don't see a use case for it. We should add it once it's needed/requested (as everything else).

``` 
const mainProjectDirectory = path.resolve(__dirname, '../../')
```

viktor-yakubiv: In this case, I think that if you do a reduce loop anyway, why not to do it more flexible to prevent such requests? But I do not insist. If you want to keep it simple, go for it.

Joozty: Also, I don't think that cosmic config should search in user home directory. But I am not sure how to make it to stop at the certain directory. Anyway, I guess it like minor thing unless somebody complains 😄

viktor-yakubiv: There is a [`stopDir` option](https://github.com/davidtheclark/cosmiconfig#stopdir) too. However, I am struggling to think the right path for it so I would agree with you 🙂

Joozty:
> I wanted to provide an interface as simple as possible when you do not need to configure all the paths and go in deep

I hope I haven't removed anything from this part. I just removed support for `output.icons` being boolean or string AFAIK. I don't think that part of the code even worked because there was wrong parameter passed to prepare function if I remember correctly. I don't think it's a big deal but I can add that support back.

viktor-yakubiv:
It's rather bug fixing that dropping a functionality 😄 Thank you

bin/config.js
Joozty: Not sure what do you mean by reduce loop. Can you maybe share some code example?

viktor-yakubiv: Sorry, I misled you and myself. Something like:

```js
const configs = []
let result
do {
  result = explorer.search(result?.filename)
  configs.push(result.config)
} while (result != null)
configs.pop() // `null` was added in the end

configs.reduce((result, current) => merge(result, current), defaultConfig)
```

but I messed up with paths, so you will need some extra conditions for handling it.

Joozty: I see. I don't think we have a use case for this. We just expect one config. So I think `explorer.search(path.resolve(__dirname, '../../'))` is more then enough.

viktor-yakubiv:
I rebased slightly adjusting first 2 commit messages and merged it to `master` to drop such simple stuff from the PR, as said before.

Also, I fixed a conflict and hope it was done properly. Sorry, if it's not.

viktor-yakubiv:
I wonder this was not picked up automatically. I had to change the base to make GitHub drop 2 commits ☝️

bin/config.js
```diff
@@ -1,41 +1,15 @@
...
-  },
-
-  svgo: {
-    plugins: {
-      removeViewBox: false,
-      removeDimensions: true,
-    },
-  },
-}
+const defaultConfig = require('../design.config')
```
viktor-yakubiv, line 6: I still do not support this. Please, restore previous lines and add it as another config into merging.

This will keep the config in place, so you do not have to go to another file to take a look at defaults (even if it's easy with IDE) and will keep your default design config more simple.

I expect this file changing rarely in comparison to `design.config.js` in the project root. It's also safer to keep that config as minimal as possible to prevent mistakes (even if we sure what we do).

design.config.js
```diff
@@ -0,0 +1,37 @@
+const path = require('path')
```
viktor-yakubiv, line 1: Since we have 2 configs here, we should add a descriptive comment on what is this particular file about.

design.config.js
```diff
@@ -0,0 +1,37 @@
...
+
+const isDev = process.env.NODE_ENV !== 'production'
```
viktor-yakubiv, line 3: Is there a way to drop this condition from the config file? I mean conditional icons adding. I am afraid it will cause problems.

design.config.js
```diff
@@ -0,0 +1,37 @@
...
+      sprite: 'icons.svg',
+    },
+  },
+
+  svgo: {
+    plugins: {
+      removeViewBox: false,
+      removeDimensions: true,
+    },
+  },
```
viktor-yakubiv, line 34: When I said, your `design.config.js` can be simpler, I meant this part is implicit. I doubt it will ever change so let's keep it out of the config and even out of the documentation (if we would have any 😁). It an internal API.

design.config.js
```diff
@@ -0,0 +1,37 @@
...
+const iconsRoot = path.join(
+  path.dirname(require.resolve('@mdi/svg/package.json')),
+  './svg'
+)
+
+const config = {
+  icons: {
+    path: iconsRoot,
+    files: icons,
+  },
```
viktor-yakubiv, line 18: I am not sure it's handled correctly. Let's imagine, we have icons in 2 packages or 2 different directories. We want to resolve one list from the first directory and another one from the second directory. Is it handled like that before merging config objects?

By the way, this an extra API I wanted to provide to the user to let them list only filenames without full paths. This is what I was talking about in the previous review.

design.config.js
```diff
@@ -0,0 +1,37 @@
+const path = require('path')
+
+const isDev = process.env.NODE_ENV !== 'production'
+
+const icons = ['ab-testing']
+
+if (isDev) icons.push('check')
+
+const iconsRoot = path.join(
+  path.dirname(require.resolve('@mdi/svg/package.json')),
```
viktor-yakubiv, line 10: Did this package become a `dependency`?

bin/config.js
```diff
@@ -54,46 +28,43 @@ const prepareIcons = (input) => {
   return files
 }
 
-const prepareOutput = (output) => {
-  const result = {
-    ...defaultConfig.output,
-    icons: { ...defaultConfig.output.icons },
-  }
-
-  if (typeof output == 'string') result.path = output
```
viktor-yakubiv, line 63: I see what you wanted to simplify. I wanted to have an ability to set the `output` to a string if we want to put everything in a single directory, e.g. when you do not want to put your icons into `public/icons` and just build everything into `public`.

It was a nice feature but I do not mind it's removed.

bin/config.js
```diff
@@ -54,46 +28,43 @@ const prepareIcons = (input) => {
...
 
-const prepare = (config) => {
-  const icons = prepareIcons(config.icons)
-  const output = prepareOutput(config.output)
-  const svgo = { ...defaultConfig.svgo, ...config.svgo }
-  return { icons, output, svgo }
+const read = async (filepath) => {
+  const explorer = cosmiconfig('design')
+  // If filepath is not supplied we start search for config
+  // from parent directory,
```
viktor-yakubiv, line 53: You say `parent` here but you actually go 2 levels above resolving `mainProjectDirectory`

bin/config.js
```diff
@@ -54,46 +28,43 @@ const prepareIcons = (input) => {
...
+
+  return mergedConfig
 }
 
-const prepare = (config) => {
-  const icons = prepareIcons(config.icons)
-  const output = prepareOutput(config.output)
-  const svgo = { ...defaultConfig.svgo, ...config.svgo }
-  return { icons, output, svgo }
+const read = async (filepath) => {
```
viktor-yakubiv, line 50: It seems you changed the function order but I do not understand why.

design.config.js
Joozty: It will at some point in https://github.com/oacore/design/pull/209

bin/config.js
Joozty: It was because of ESlint warning that function was used before it is declared.

design.config.js
Joozty: I don't know. I mean I didn't like it either but that was the only solution I could find.

design.config.js
viktor-yakubiv: It's better to list this as s `dependency` here since you use `path.resolve` to it in the config.

design.config.js
viktor-yakubiv: I think that we should build a custom library version for the documentation, i.e. build the documentation with the library as a dependency and have a separate configuration there (with icons needed for the documentation examples).

However, using some CDN seems to be even better to me because it may work even it markdown.

Let's keep this out of scope at the moment.

design.config.js
Joozty: That's config is meant to be removed before merge

design.config.js
Joozty: So it's possible to use different icon package in design and different icon package in dashboard but it's not possible to use icons from different packages in the same project unless you provide full paths. It's not ideal however we don't need such a config at the moment.

bin/config.js
Joozty: If it makes you happy why not. I don't see much difference.

design.config.js
Joozty: Will be added in next PR.

design.config.js
viktor-yakubiv: That is why I wanted to have a custom merge function: to merge half-prepared configs and then finish expanding all the parameters. I assume this would fix the issue we have here.

I understand it's not needed at the moment, so I resolve the comment but I will create another issue for this.

design.config.js
viktor-yakubiv: Sorry, it was clearly said in the commit message, I forgot about. Sure, let's do it in #209.

bin/config.js
viktor-yakubiv: I was thinking about it a bit more. It comes from the single responsibility principle (SRP). Let's imagine we have many developers of the library. Some of them are designers, who work only on the components and some are engineers who work on the automation part. Engineers don't need to add icons but provide the API to designers who add the components and the icons. The designers, at the same time, should not care about the defaults if there is anything missing but only want to have clear and simple config.

Joozty: All right that makes sense. Thanks for explanation. In that case we shouldn't allow to override svg config at all, should we? I mean it's really minor and it can stay as is.

design.config.js
Joozty: sure, it's good to have it in backlog

bin/config.js
viktor-yakubiv: Yes, this assumption seems to be correct but I would keep this possibility at the moment as it's minor 🙂

Maybe we will need it in the future.

PR #207, opened 2020-07-07
Add flex layout to the FormGroup

viktor-yakubiv:
It's a dangerous thing that is required for the following steps in development but may **break** other product. Please, take a look at the comment in the file.

### DIFF
--- a/src/elements/forms/styles.css
+++ b/src/elements/forms/styles.css
@@ -27,6 +27,7 @@
 
 %label {
   display: block;
+  flex-grow: 1;
   margin-bottom: 0.5rem;
   line-height: 1;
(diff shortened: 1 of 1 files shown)


viktor-yakubiv:
I do not understand why we have an extra job in this PR which does not pass.

Joozty:
LOL, isn't it that thing you tried few months ago? Looks like they have still the access to the repository. I think it was this https://www.codacy.com/

viktor-yakubiv:
I checked the setting but there is nothing, I am afraid. It may be a bug or something

viktor-yakubiv:
Nothing should change. That's all

PR #208, opened 2020-07-07
Develop InputAddon component

viktor-yakubiv:
I am struggling to make the extra text transparent when the label is shown at the moment.

### DIFF
--- a/src/elements/forms/forms.md
+++ b/src/elements/forms/forms.md
@@ -44,8 +44,8 @@ import ProgressSpinner from '../progress-spinner';
   <FormAddon>
     <ProgressSpinner />
   </FormAddon>
-  <FormControl type="number" id="example-input-2" placeholder="e.g. 10" />
-  <FormLabel htmlFor="example-input-2">Amount</FormLabel>
+  <FormControl type="text" id="example-input-3" placeholder="e.g. 10" />
+  <FormLabel htmlFor="example-input-3">Amount</FormLabel>
 </FormGroup>
 ```
 
--- a/src/elements/forms/forms.md
+++ b/src/elements/forms/forms.md
@@ -21,10 +21,31 @@ Label after the input:
 import FormLabel from './label';
 import FormControl from './control';
 import FormGroup from './group';
+import FormAddon from './addon';
 
 <FormGroup>
-  <FormControl placeholder="Placeholder" />
-  <FormLabel>Label</FormLabel>
+  <FormAddon place="prepend">$</FormAddon>
+  <FormAddon place="append">.00</FormAddon>
+  <FormControl type="number" id="example-input-2" placeholder="e.g. 10" />
+  <FormLabel htmlFor="example-input-2">Amount</FormLabel>
+</FormGroup>
+```
+
+Input with a progress indicator:
+
+```jsx
+import FormLabel from './label';
+import FormControl from './control';
+import FormGroup from './group';
+import FormAddon from './addon';
+import ProgressSpinner from '../progress-spinner';
+
+<FormGroup>
+  <FormAddon>
+    <ProgressSpinner />
+  </FormAddon>
+  <FormControl type="number" id="example-input-2" placeholder="e.g. 10" />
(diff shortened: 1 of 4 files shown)


viktor-yakubiv:
The ProgressSpinner is almost there! Like 2 pixels extra 😅

viktor-yakubiv:
I was exploring these 2 extra pixels and it's a painful thing. It comes from nowhere, browsers just add it (Firefox and Chrome tested). It should be explored in deep to be properly fixed or a workaround should be added for now. @Joozty what do you think?

viktor-yakubiv:
Also, the top border is flickering because of the label width changing instantly. I think that transition will not make it better but also I do not know how to fix it.

src/elements/forms/forms.md
```diff
@@ -21,10 +21,31 @@ Label after the input:
...
+```jsx
+import FormLabel from './label';
+import FormControl from './control';
+import FormGroup from './group';
+import FormAddon from './addon';
+import ProgressSpinner from '../progress-spinner';
+
+<FormGroup>
+  <FormAddon>
+    <ProgressSpinner />
```
Joozty, line 45: You will need to make progress spinner display flex or block otherwise the height of it doesn't apply since height and width don't work on inline elements.

Alternatively, you can make FormAddon display flex/block and it should fix it too.

Joozty:
> Also, the top border is flickering because of the label width changing instantly. I think that transition will not make it better but also I do not know how to fix it.

Can you elaborate? When is the top border flickering? I don't see it.

src/elements/forms/forms.md
viktor-yakubiv: I am considering adding flex to FormAddon as a solution but I am afraid it will behave unpredictable for the user.

`ProgressSpinner` should be inline-block, is not it? It's applied transitionally by `svg` selector from `sanitize.css`.

viktor-yakubiv:
When you click on the input and click outside without typing any text, the label width changes and the label with white background transits from the top to centre overlapping the border. It's what I described 'flickering'.

src/elements/forms/forms.md
Joozty: This is maybe a good question for StackOverflow. If you aren't going to ask I will. I started going deeper and I am not even sure if I was right in the previous comment. These 2.6 pixels is a mystery.

viktor-yakubiv: I was looking for some resources. Although, if you want to you can ask it on StackOverflow. I am pretty sure there should an answer already but it may be tricky to find 🙂

Joozty: So I asked [here](https://stackoverflow.com/a/62803300/5594539) and based on the answer I would still say that display: block is the best for us. I will keep it up to you. 


OT:
Maybe we should use no-wrap [here](https://github.com/oacore/design/blob/forms/src/elements/forms/styles.css#L226). When the input is small it is broken to two lines. I should have noticed before. Did you have a reason for having it wrap?

viktor-yakubiv: I will try inline-flex.

I need wrapping because I use the same element for everything: label, input group and helper text. I do not like [Bootstrap's solution](https://getbootstrap.com/docs/4.5/components/input-group/) where another wrapped is needed for `input-group` to work. I understand that it's safe but I would rather put more pressure onto the developer than add some unnecessary wrapper.

In the same time, the developer can add another wrapper if there is any problem with automatic resizing. I would be happy if you help me to document this properly.

Also, we can make TextField safer to use adding such wrapper. However, the low-level API is open to discussion.

viktor-yakubiv: `inline-flex` seems to be an option, thank you.

However, it does work as expected because of the default `flex-direction: row`. However, when you put there something like `<Icon /> some text` (pay attention to spaces), you lose the space between the icon and the text line because the text node is converted into anonymous flex block and space gets reduced. This is desired behaviour often but this is not the default  HTML behaviour and we must keep it in mind.

Off-topic, JSX suffers from the same problem.

Issue #205, opened 2020-07-03
Develop a Link component


viktor-yakubiv:
Add a wrapper around the `<a>` element. The interface should be similar to [Vercel's Link](https://vercel.com/design/link):

```jsx
<Link href="https://core.ac.uk" external>CORE</Link>
```

I propose to support only `external` prop at the moment. The icon for the `external` links should probably be added always but let's not do it later because there is a caveat: what if the link is wrapped around a _Card_ or any other element.

We can consider other props later. However, it will probably be `noIcon`. The colour should be always overridden by CSS, we are not going to support it ever.

viktor-yakubiv:
I think that we can support the `variant` for the Link. A variant can add an icon or slightly change the look. However, it makes it quite complex when the Button becomes dependent on this component.

Joozty:
@viktor-yakubiv do you want to make button dependent on Link? I would say these elements should be separated. And even if users use `<Button href="/" tag="a" />` the button component should behave as is now. Do you have a different point of view? 🙂

viktor-yakubiv:
I think that the `tag` prop should not be allowed for both Button and Link:
1. the Link always is `<a>`
2. the Button detects the tag based on the `href` prop: `href != null ? Link : 'button'`.

I do not like such dependency but I was thinking that this way is more useful.

From the second point, we see that the button cannot be an `<input>`. It's more intentional, I do not mind removing such support since we assume having children always. Also, I was thinking about `<Button tag="span" />`: this is semantically wrong and should not be allowed too.

PR #209, opened 2020-07-09
Add Link component

Joozty:
Closes https://github.com/oacore/design/issues/205


src/elements/link/index.js
```diff
@@ -0,0 +1 @@
+export default from './link'
```
viktor-yakubiv, line 1: I do not like this convention. Do you see any other better option to handle component modules?

src/elements/link/link.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Link = React.forwardRef(
+  ({ children, className, external = false, ...restProps }, ref) => (
+    <a
+      ref={ref}
+      className={classNames.use(styles.link).join(className)}
+      {...restProps}
+    >
+      {children}
+      {external && <Icon src="#iconhere" alt="external" />}
```
viktor-yakubiv, line 17: The reason for `external` is not in the Icon, it's fine without, I would rather drop it. The prop is a shortcut for `target="_blank" rel="noopener noreferrer"`

Joozty: alt is required prop. I just put something in there.  So you are saying we should drop required alt prop?

src/elements/link/index.js
Joozty: Not at the moment. :(

src/elements/link/link.jsx
viktor-yakubiv: I do, we should drop the requirement from the `Icon` but it's slightly different, out of this scope.

In the previous comment, I mean that the `external` prop should add `target="_blank"` but the icon itself is optional.

Joozty: Thanks will add it. It still work in progress so I just put blindly things here.

Joozty:
@viktor-yakubiv Don't you know by any chance the name of the external icon? It's like finding a needle in the haystack.

viktor-yakubiv:
If I remember, `open-in-new`

viktor-yakubiv:
> If user click on Link in menu they are redirected to the button link instead because it has the same ID.

The backside of the fancy markdown 😄

viktor-yakubiv:
I am looking into it. I will send an update, I am exploring more cases. However, I like the overlall result. Thanks a lot for doing this 👍

viktor-yakubiv:
I found the icon looks good in Safari but in Chrome and Firefox it does not look very good in text. Feel free to check it now but I will apply some changes.

By the way, I do not recommend to change the default display for the text elements, it may be unexpected for the user.

viktor-yakubiv:
There is a problem that I see can only be solved with web-fonts at the moment (without JS processing). I wish I am wrong and there is an option with the background + mask bit obviously inline-block behaves like an image too.

Open in new icon falls to the next line

The icon should break with the last work within the link alway. `white-space` does not help us because it will make the whole link unbreakable.

viktor-yakubiv:
I added a simple solution but we should investigate this breaking rules more because it will become a headache when goes to production.

viktor-yakubiv:
@Joozty sorry I put my hands here maybe too much. I tried to explain my concerns as much as possible and even created an issue after this (#216). I wonder if you have any comments but if you don't feel free to rebase and merge 😉

Thanks a lot for looking into this. It was interesting investigation, I hope for both of us 🙂

Joozty:
> The icon should break with the last work within the link alway. white-space does not help us because it will make the whole link unbreakable.

You would need another wrapper I guess.
```
<a>Link with the last <span>word<svg>icon here</svg></span></a>
```

Anyway thanks for looking into this. As I always say I don't see any difference whether the element is inline flex or just inline. 😄

viktor-yakubiv:
This is exactly the solution I was thinking about but it needs language processing on JS what I would like to avoid. I would rather have this defect than handle it in such way.

PR #211, opened 2020-07-09
Add Message component

Joozty:


Joozty:
@viktor-yakubiv what do you think Do we need additional styling or is it ok like that?

src/elements/message/message.css
```diff
@@ -0,0 +1,26 @@
+:root {
+  --message-success-color: var(--success);
+  --message-warning-color: var(--warning);
+  --message-danger-color: var(--danger);
```
viktor-yakubiv, line 4: Use dark accents here please

src/elements/message/message.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import styles from './message.css'
+
+import { classNames } from 'utils'
+
+const Message = React.forwardRef(
+  ({ children, className, variant, tag: Tag = 'span', ...restProps }, ref) => (
```
viktor-yakubiv, line 9: Tag should be <p> probably.

PR #212, opened 2020-07-09
Fix React warnings

Joozty:


styleguide.config.js
```diff
@@ -34,7 +34,6 @@ module.exports = {
             'src/elements/table/body.jsx',
             'src/elements/table/row.jsx',
             'src/elements/table/cell.jsx',
-            'src/elements/table/body.jsx',
```
viktor-yakubiv, line 37: Why is this changed?

Joozty: it's duplicated. The body is declared three lines up. This duplicate value causes warning about the same key. (not sure why really)

PR #217, opened 2020-07-16
Add CORE logo component

Joozty:
Closes https://github.com/oacore/design/issues/49


src/elements/logo/logo.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+
+import Icon from '../icon'
+import styles from './logo.css'
+
+import { classNames } from 'utils'
+
+const Logo = React.memo(
+  ({
+    children = 'Dashboard',
```
viktor-yakubiv, line 10:
```suggestion
    children = 'CORE',
```

src/elements/logo/logo.jsx
```diff
@@ -0,0 +1,24 @@
...
+    className,
+    src = '#core-symbol',
+    alt = 'CORE',
```
viktor-yakubiv, line 13: I think it's better to have it editable as it is but I am not sure. What's your opinion?

src/elements/logo/logo.md
```diff
@@ -0,0 +1,7 @@
+# CORE logo
+
+
+```jsx
+<Logo>Design</Logo>
```
viktor-yakubiv, line 5: Please, add an example of usage with _Link_, it's more popular to have the logo in `<a>`

src/elements/logo/logo.jsx
Joozty: It allows to use different logos, on the other hand, we apply custom negative margin on icon so any other icon can look badly 🤷‍♂️

viktor-yakubiv: Let's keep it 🙂

PR #220, opened 2020-07-16
Text field adjustments

viktor-yakubiv:


viktor-yakubiv:
I am struggling with the API. When I started using my components, I see it does not work as I want. Do you have any requirements or ideas, @Joozty?

Joozty:
I don't have any requirements or ideas. You are on a good path.  I wanted to include progress in textfields and you already did it. :)

viktor-yakubiv:
I don't like that the `status` prop basically repeats the `variant`. I am struggling with combining colour and shape:
- the colour is controlled by the `variant`
- the icon look is controlled by the `status`

Technically, it's possible to use `variant="error" status="valid"` what is not correct.

I am thinking about making the `variant` even more powerful, allowing to pass multiple options separated by a space. I see many advantages for that but this is out of the scope.

Joozty:
what if you would make icon completely depended on variant and you would add another `icon` prop which could be boolean/string?

True - Show default icon for variant
False - Don't show any icon
String - show specified icon

viktor-yakubiv:
I don't like the interface much and may change it later to pure `variant` unless you have arguments against it.

Also, I am afraid the `ProgressSpinner` component is too heavy. When the documentation is open in any browser, my laptop shows that the browser uses significant energy and the cooler starts working hard. I got the animation from the internet and it does not seem to be so heavy to me. I wonder if there is any way to optimise that but I have no idea how and only hope it does not affect the performance for short amounts of time.

PR #223, opened 2020-07-21
Optimise spinner animation

Joozty:


viktor-yakubiv:
I was reading about `will-change`, it's not recommended to use or is recommended to remove after the animation because it drains resources too.

If we have not done it yet we can uniform transform properties and maybe leave `will-change` for stroke only. What do you think?

Joozty:
What do you mean by uniforming transform properties?

viktor-yakubiv:
I meant that we should indicate `transform: rotate(0)` even if there is no need to change so the browser knows there should be rendered a separate layer.

viktor-yakubiv:
However, your solution is quite good too. Maybe there is no need to change. If you wich you can test it once more if you don't, feel free to merge. Thanks a lot 👍

Joozty:
It's already in separate layer whether there is transform: rotate(0) or will-change. However, I observed that will-change reduced slightly CPU Usage. I don't have a big knowledge in these things maybe I measured it wrong or it's a coincidence.

viktor-yakubiv:
Don't worry about this much. It's already good we explored this and know where to dig in 😄

If we need many loading icons I will design more simple way to indicate this. I have a few ideas already.

Thanks a lot for doing this 👍

PR #225, opened 2020-07-22
Fix flickering animation

Joozty:


Joozty:
Looks like Safari is playing a game with me. Firefox and Chrome work like a charm. :(

Joozty:
@viktor-yakubiv what do you think about this workaround? I am not sure if there is a way to fix that right padding in safari.

Joozty:
Negative delay skips the first Xms of animation. The animation is not that smooth as before but it's not bad either.

src/elements/forms/styles.css
```diff
@@ -34,7 +34,7 @@
   line-height: 1;
   color: var(--form-label-color);
   border-radius: var(--form-control-corner-radius, 2px);
-  transition: 150ms;
+  transition: all 150ms ease -50ms;
```
viktor-yakubiv, line 37: Ease is the default value, there was no need to add it.

PR #227, opened 2020-07-22
Fix label

Joozty:


viktor-yakubiv:
I guess I removed `overflow-x` from the FormGroup but I forgot about `.label::before` so it may break something. Let me know if I should take a look at it too.

Joozty:
I mean if you want to and you kno whow to remove this ::before selector feel free to have a look. If you don't have a time let me know and I will merge it and release. (The issue I ran into is solved by this change, so I guess it's ok as is)

viktor-yakubiv:
We can release another patch with my changes. Feel free to release `0.3.1` for now 🙂👍

PR #234, opened 2020-07-23
Make alt for Icon optional & allow to pass element as a helper prop for TextField

Joozty:
We are missing alt prop for addon icons in [TextFields](https://github.com/oacore/design/blob/master/src/elements/forms/text-field.jsx#L31-L39). I don't think it makes sense to put there any alt attribute because we use arria-hidden there. If you have a better solution for this problem let me know.

### DIFF
--- a/src/elements/forms/text-field.jsx
+++ b/src/elements/forms/text-field.jsx
@@ -134,7 +134,7 @@ TextField.propTypes = {
   /**
    * Helper message to show bellow the input
    */
-  helper: PropTypes.string,
+  helper: PropTypes.node,
   /**
    * Helper variation. Affects label position and the colour.
    *
--- a/src/elements/icon/icon.jsx
+++ b/src/elements/icon/icon.jsx
@@ -19,9 +19,9 @@ const Icon = ({ src, alt, className, ...restProps }) => (
   <svg
     className={classNames.use(iconClassNames.icon).join(className)}
     role="img"
+    aria-label={alt}
     {...restProps}
   >
-    <title>{alt}</title>
     <use href={resolve(src)} />
   </svg>
 )
@@ -37,7 +37,7 @@ Icon.propTypes = {
   /**
    * Alternative text for accessibility
    */
-  alt: PropTypes.string.isRequired,
+  alt: PropTypes.string,
 }
 
 export default Icon
--- a/src/elements/icon/icon.md
+++ b/src/elements/icon/icon.md
@@ -1,8 +1,8 @@
 ## Accessible icon
 
-Provide alternative text in `alt` prop for tech assistive users accessibility.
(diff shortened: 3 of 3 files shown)


src/elements/icon/icon.jsx
```diff
@@ -21,7 +21,7 @@ const Icon = ({ src, alt, className, ...restProps }) => (
     role="img"
     {...restProps}
   >
-    <title>{alt}</title>
+    {alt && <title>{alt}</title>}
```
viktor-yakubiv, line 24: We can keep this for the consistency but it should be changed:
- [ ] remove the title at all
- [ ] add `alt` as `aria-label` to the SVG
- [x] make it optional

`<title>` works as the `title` attribute, it shows a tooltip and this is not what we want. Alt is only for the accessibility when the user cannot see the image, but only can read the text; to my mind, `aria-label` fits it better.

src/elements/forms/text-field.jsx
```diff
@@ -134,7 +134,7 @@ TextField.propTypes = {
   /**
    * Helper message to show bellow the input
    */
-  helper: PropTypes.string,
+  helper: PropTypes.oneOfType([PropTypes.element, PropTypes.string]),
```
viktor-yakubiv, line 137: Would not `PropTypes.node` define these 2?

src/elements/forms/text-field.jsx
```diff
@@ -31,7 +31,7 @@ const Status = ({ value, variant, progress, ...restProps }) => (
       <Icon src={value} aria-hidden />
     ) : (
       {
-        valid: <Icon src="#check" aria-hidden />,
+        valid: <Icon src="#check" alt="valid icon" aria-hidden />,
```
viktor-yakubiv, line 34: Don't use `alt` at all.

Joozty: Oh pushed by mistake.

src/elements/icon/icon.jsx
Joozty: You confused me with the checked box. I thought you want to leave it as an optional and solve is later.

viktor-yakubiv: Sorry, I just liked this type of list a bit more 🙂

Also, you can mark it If I am not mistaken.

src/elements/icon/icon.jsx
```diff
@@ -19,9 +19,9 @@ const Icon = ({ src, alt, className, ...restProps }) => (
   <svg
     className={classNames.use(iconClassNames.icon).join(className)}
     role="img"
+    aria-label={alt}
```
Joozty, line 22: @viktor-yakubiv Is it how you imagined it? If so please merge.

PR #240, opened 2020-07-30
Add automatically displayName to functional components

Joozty:
This helps to debug errors in production build and also helps to distinguish children by type,e.g. `children.type.displayName = 'SelectOption'`

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -18,6 +18,7 @@ module.exports = {
       {
         alias: {
           utils: './src/utils',
+          elements: './src/elements',
         },
       },
     ],
@@ -27,5 +28,6 @@ module.exports = {
     '@babel/plugin-proposal-object-rest-spread',
     '@babel/plugin-transform-runtime',
     ['css-modules-transform', cssModulesConfig],
+    'add-react-displayname',
   ],
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -3770,6 +3770,11 @@
         }
       }
     },
+    "babel-plugin-add-react-displayname": {
+      "version": "0.0.5",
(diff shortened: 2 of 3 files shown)


Joozty:
Well it's about finding correct chidlren. Imagine something like this:

```
<Select>
    <Select.Option>1</Select.Option>
    <Select.Option>2</Select.Option>
    <Select.Option>3</Select.Option>
    <Select.Option>Search all results</Select.Option>
    <Select.Value>
        <div>Eleement on top of input when not focused</div>
     </Select.Option>
</Select>
```

You may want to find all options and get props from them, then you want to find Section.Value component and check if there is only one of its type (if not throw exception). We use this approach in [Dashboard Table]() too.  I don't see anything wrong with it...  :) 


Note the example doesn't reflect the reality. I probably won't use <Select.Value/> (at least not in the first version)

viktor-yakubiv:
> I probably won't use <Select.Value/> (at least not in the first version)

Please, don't 😄

viktor-yakubiv:
You may enforce it, this is where you go. However, I would let the user pass whatever he/she/they want to but unexpected behaviour is also on their shoulders.

In my example, if you have

```jsx
<Select value="a value">
  <Menu.Item value="a value">A value</Menu.Item>
</Select>
```

it may behave correctly but it would not work with the filtering you want.

Joozty:
The rationale behind it is to allow custom styling and reuse this component in the dashboard. Something like (not needed at the time, it was more just to provide you example):

```
    <Select.Value>
        <bold>Org name</bold>
        <span>Repo name</div>
     </Select.Value>
```

Joozty:
Your example is just fine. So it will be just one condition more...

PR #270, opened 2020-08-05
Refactor Select interface

Joozty:
Changes:
 - onInput callback - called when input changes
 - onChange callback - called when user hits enter or select suggestion
 - initialValue - user is now able to specify only initial input value the value changes are handle internally and user may subscribe for it via onInput callback


Joozty:
I drop support for controlled component for some reason which I don't remember now. (not great since it's been just a few hours since I developed it 😄 ) Anyway, I guess it can be added later on as we need it. For not the solution should be sufficient.

PR #273, opened 2020-08-05
Refactor select vol2

Joozty:


src/modules/select/hooks/use-input.js
```diff
@@ -1,17 +1,17 @@
...
   const inputRef = useRef(null)
-  const [inputData, changeInputData] = useState({ value: initialValue })
+  const [inputData, changeInputData] = useState({ value })
   const [isInputFocused, setIsInputFocused] = useState(null)
 
   useEffect(() => {
-    changeInputData({ value: inputData.value })
-  }, [initialValue])
+    if (value !== inputData.value) changeInputData({ value })
+  }, [value])
```
viktor-yakubiv, line 10: Why don't you blindly pass it down?

Joozty: To prevent calling that callback several times with the same value

PR #287, opened 2020-08-17
Drop generateId support

Joozty:
This never really worked for SSR applications because different ID is generated on server side than on client side. There are some ways to make it work properly but it looks too complicated. Let the user pass it down.

See more:
https://github.com/facebook/react/issues/5867


viktor-yakubiv (approved):
It's a good move! Thank you 👍

PR #289, opened 2020-08-19
Add Popover component

Joozty:
Closes https://github.com/oacore/design/issues/290

### DIFF
--- a/src/modules/popover/popover.md
+++ b/src/modules/popover/popover.md
@@ -1,0 +1,107 @@
+```jsx
+
+import TextField from '../../elements/forms/text-field';
+import Button from "../../elements/button";
+
+
+<div
+  style={{
+    margin: '5rem',
+    display: 'grid',
+    gridTemplateColumns: 'repeat(3, 1fr)',
+    gridGap: '0.5rem',
+    gridAutoRows: 'min-content',
+  }}
+>
+  <Popover
+    placement="top"
+    content="I am a simple tooltip with placement='top'"
+  >
+    <Button variant="outlined" style={{ width: '100%' }}>
+      Top
+    </Button>
+  </Popover>
+  <Popover
+    placement="top-start"
+    trigger="click"
+    content={
+      <ul>
+        <li>
+          <code>placement</code> top-start
+        </li>
+        <li>
+          <code>trigger</code> click
+        </li>
+      </ul>
+    }
+  >
+    <Button variant="outlined" style={{ width: '100%' }}>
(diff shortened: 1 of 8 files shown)


src/modules/popover/content.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+import Card from 'elements/card'
+import { classNames } from 'utils'
+
+/**
+ * @visibleName Popover.Content
+ */
+const PopoverContent = forwardRef(
+  ({ children, className, ...passProps }, ref) => (
+    <Card
```
viktor-yakubiv, line 13: The `Card` might be too heavy for the tooltips. It's better to create something else even if it looks similar.

src/modules/popover/popover.md
```diff
@@ -0,0 +1,155 @@
...
+        <Popover.Content>
+          I am a simple tooltip with placement='top'
+        </Popover.Content>
+        <Popover.Trigger
+          tag={Button}
+          variant="outlined"
+          style={{ width: '100%' }}
+        >
+          Top
+        </Popover.Trigger>
```
viktor-yakubiv, line 26: The interface seems to be complicated to me. I like how it's designed but I am not sure you or myself will be happy to use this many times.

This can be used internally but the final interface is desired to be more simple, similarly to `TextField` and `Form.*`. What do you think?

viktor-yakubiv (approved):
I like it overall 👍

However, I am afraid that the component is too powerful: it allows to show basic tooltips and rich menus, it works on hover, focus, click. I feel like users, even ourselves and especially the developers after us will start using it in a wrong way: rich menu on hover or bare text on click. This is not what I would like to have, so I recommend to provide a few more specific APIs based on or instead of this generic one.

src/modules/popover/popover.md
Joozty: Do you have any suggestions?

What about this one?

```jsx
<Tooltip
  content="either text or react lement"
>
<button>Click me!</button>
</Tooltip>
```

viktor-yakubiv: I like it but not completely. We might want to support rich formatting. In this case, the proposed tooltip will become bloated

```jsx
<Tooltip
  content={(
    <>
      <Icon src="#alert">
      <h3>Error happened<h3>
      <p>Who knows, what is this<p>
    </>
  )}
>
  <button>Click me!</button>
</Tooltip>
```

I propose something slightly different in #290. Let's discuss it there.

PR #293, opened 2020-08-25
Move sanitize css to extra file

Joozty:
The user may easily import just this reboot file and typography. Useful for migrations (@oacore/next)


viktor-yakubiv:
I am not sure about the change but it's better than it was for sure.

I would prefer to import sanitize.css contextually, i.e. where it's needed. Of course, I may be wrong because I don't clearly imagine what I think about.

Thank you 👍

PR #297, opened 2020-08-28
Filter Select children

Joozty:
Allows to use this kind of syntax:

```
const booleanValue = false
<Select>
  {booleanValue && <Select.Option />}
</Select>
```

### DIFF
--- a/src/modules/select/hooks/use-options.js
+++ b/src/modules/select/hooks/use-options.js
@@ -22,34 +22,39 @@ const useOptions = (
   }
 
   const options =
-    React.Children.map(children, (element) => {
-      const overrideProps = {
-        'id': `suggestion-result-${selectId}-${element.props.id}`,
-        'value': element.props.value,
-        'data-select-id': element.props.id,
-        'onMouseDown': () => {
-          // Firefox and Safari have relatedTarget null.
-          // Therefore we cannot use onClick event
-          // since it's not getting fired
-          // because onBlur input listener cause rerender.
-          setClickedElement(element.props.id)
-        },
-        'onMouseUp': (event) => {
-          if (clickedElement === element.props.id) {
-            changeActiveOption({
-              value: element.props.value,
-              id: element.props.id,
-            })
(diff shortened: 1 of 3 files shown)


src/modules/select/hooks/use-options.js
```diff
@@ -22,34 +22,39 @@ const useOptions = (
...
-        },
-        'onMouseUp': (event) => {
-          if (clickedElement === element.props.id) {
-            changeActiveOption({
-              value: element.props.value,
-              id: element.props.id,
-            })
-          }
+    React.Children.toArray(children)
+      .filter((child) => child.type.displayName === 'SelectOption')
```
viktor-yakubiv, line 26: What if the `child` is `null`? Is it filtered out already by `React.Children.toArray`?

src/modules/select/hooks/use-options.js
```diff
@@ -22,34 +22,39 @@ const useOptions = (
...
+        const overrideProps = {
+          'id': `suggestion-result-${selectId}-${element.props.id}`,
+          'value': element.props.value,
+          'data-select-id': element.props.id,
+          'onMouseDown': () => {
+            // Firefox and Safari have relatedTarget null.
+            // Therefore we cannot use onClick event
+            // since it's not getting fired
+            // because onBlur input listener cause rerender.
+            setClickedElement(element.props.id)
```
viktor-yakubiv, line 37: Is it actually clicked, hovered, focused or selected?

viktor-yakubiv (approved):
It's a good improvement!

src/modules/select/hooks/use-options.js
Joozty: Looks like it is indeed

src/modules/select/hooks/use-options.js
Joozty: It's an element which received onMouseDown event. If the same element receives onMouseUp event then it is considered to be clicked.

PR #298, opened 2020-08-28
Add prop changeOnBlur

Joozty:
If set to false the onChange is not called on blur event

### DIFF
--- a/src/modules/select/hooks/use-input.js
+++ b/src/modules/select/hooks/use-input.js
@@ -1,9 +1,10 @@
-import { useEffect, useRef, useState } from 'react'
+import { useEffect, useRef, useState, useCallback } from 'react'
 
-const useInput = (value, { onInput, onChange }) => {
+const useInput = (value, { onInput, onChange, changeOnBlur }) => {
   const inputRef = useRef(null)
+  const inputEventType = useRef(null)
   const [inputData, changeInputData] = useState({ value })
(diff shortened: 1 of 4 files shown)


Joozty:
@viktor-yakubiv I don't like the prop name, do you have anything better to suggest?

Joozty:
@viktor-yakubiv ping :)

viktor-yakubiv (approved):
I don't understand what you actually are doing here and have no suggestions to the better name.

If this is a search field with autosuggestions, I imagine this component to be a custom replica of the `<input>` + `<datalist>`. Take a closer look at how the [datalist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) component works.

I see that:
- when you type, it filters the results
- when you press arrows it navigates through the suggestions but the value stays unchanged
- when you click outside, the value remains unchanged
- when you select the value (using 'Enter' or mouse') the input value changes

If we use such native compoent, there are two ways to detect what option is selected:
1. dangerously check the new value (which is textual)
2. introduce an event like `onSelect` (better) or `onChoose` that is triggered when the option is selected. If it's prevented, the selection is prevented and the value remains the same.

If we have such an event, we can rely on it and do some `history.push()` if the option should navigate to the page but not only change the value of the input.

This requires 2 layers of abstraction (to add one more) but it feels right to me. Does this make any sense to you?

---

However, if you need to solve a smaller problem right now, I approve what you did here. It looks good to me too 🙂

Issue #322, opened 2020-09-11
Class names discovery from the context ✨


viktor-yakubiv:
I’ve just come to the idea that we can simplify our components API if we utilise context technique in the library on the top level. For the instance, we have:

```jsx
<Card>
  <Card.Title>I'm a card!</Card.Title>
</Card>
```

that turns into

```html
<div class="card">
  <h2 class="card-title">I'm a card!</h2>
</div>
```

## The proposal

Potentially, we can have a bit easier API like

```jsx
<Card>
  <Heading>I'm a card!</Heading>
</Card>
```

that produces the same result.

## Simple implementation

```jsx
const SectionContext = React.createContext({ level: 1 })
const ClassNamesContext = React.createContext({})

const Card = ({ children, className, tag: Tag = 'div', ...restProps }) => (
  <SectionContext.Provider value={{ level: 2 }}>
    <ClassNamesContext.Provider value={{ heading: 'card-title' }}>
      <Tag className={joinClasses('card', className)} {...restProps}>
        {children}
      </Tag>
    </ClassNamesContext.Provider>
  </SectionContext.Provider>
)

const Heading = ({ children, className, ...restProps }) => {
  const level = useContext(SectionContext)?.level ?? 1
  const contextClassName = useContext(ClassNamesContext)?.heading

  const Tag = `h${level}`

  return (
    <Tag className={joinClasses(contextClassName, className)} {...restProps}>
      {children}
    </Tag>
  )
}
```

## Use cases

My use case above is pretty simple and seems to be not worth it. I not sure either in this.

Apart from `Heading` replacing the `Card.Title` along with `Section.Title` we may have in the future, it looks more convenient for implicit passing classes to:
- `Icon` that is descendent of the `TextField`, `Select` or `Menu`
- similar to `Button` that is descendet of the `TextField`

I am pretty sure, there are other cases. However, since it's implicit, it's not obvious: very easy to use, very hard to debug.

@Joozty what do you think?

viktor-yakubiv:
This can be also scaled for the headers accessibility. Check the example:

```jsx
<Section id="my-section">
  <Heading>My section</Heading>
</Section>
```

should produce:

```html
<section id="my-section" aria-labelledby="my-section-title">
  <h2 id="my-section-title">My section</h2>
</section>

viktor-yakubiv:
However, we should investigate React Context API performance. If it's bad, we should drop the idea, unfortunately. Especially interesting it the performance of combining many contexts.

[...]

PR #327, opened 2020-09-18
Add Section and SectionHeading

viktor-yakubiv:
I am not sure this is work in progress or can be merged. The implementation is copied from oacore/dashboard#443. 

Styles are not copied, I think these should be separated. The rationale behind is that the current implementation allows doing something like this:

```jsx
<Page id="ref">
  <Heading>REF 2021</Heading>
  <Card id="overview" tag={Section}>
    <Heading>Overview</Heading>
  </Card>
</Page>
```

when the styled implementation will disable such functionality. However, I am not sure about such compositing either.

This partially relates to #322 but I find that class discovery should be avoided as much as possible when `aria-labelledby` is very useful.

### DIFF
--- a/src/content/context.js
+++ b/src/content/context.js
@@ -1,0 +1,11 @@
+import { createContext } from 'react'
+
+const defaultContext = {
+  level: 1,
+  parentId: null,
+}
+
+const SectionContext = createContext(defaultContext)
+
+export default SectionContext
+export { defaultContext }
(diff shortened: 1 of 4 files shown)


Joozty (approved):
I wouldn't mind if the styles are here too but it's up to you.  Feel free to merge once you solve the error in the pipeline.

viktor-yakubiv:
This is actually pretty need work, I can finish this if you would need it. @vkot91 what would you say?

viktor-yakubiv:
I find this stale and not needed. It was not bad chunk of work to be honest 🙂

PR #345, opened 2020-10-05
Elevation and colours based on the Material Design

viktor-yakubiv:
@Joozty, these are minor changes that are not used yet but would be very useful in the future. I am afraid I will have no time to move this forward in the nearest future, so I propose to merge it as is if you don't mind.


viktor-yakubiv:
Yeah, and I want to disable opinionated Prettier. I am opinionated enough to not use somebody's else opinion.

Joozty (approved):

PR #347, opened 2020-10-06
Bump @oacore/stylelint-config-base to 2.0.0-stylelint.1

viktor-yakubiv:
Drops Prettier support in the Stylelint.

Updates CSS files after rule changes in the stylelint config.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "@babel/preset-react": "^7.10.4",
     "@oacore/eslint-config-react": "^1.0.6",
     "@oacore/prettier-config": "^1.0.6",
-    "@oacore/stylelint-config-base": "^2.0.0-stylelint",
+    "@oacore/stylelint-config-base": "^2.0.0-stylelint2",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "@babel/preset-react": "^7.10.4",
     "@oacore/eslint-config-react": "^1.0.6",
     "@oacore/prettier-config": "^1.0.6",
-    "@oacore/stylelint-config-base": "^1.0.6",
+    "@oacore/stylelint-config-base": "^2.0.0-stylelint",
```
Joozty, line 24: You should have kept just 2.0.0 without the suffix.

viktor-yakubiv: What's the problem with the prerelease tag?

Joozty: Nothing is wrong but now all @oaocore/configs packages will need to be upgraded to this version with prerelease tag and it doesn't look good to my eye 😄

viktor-yakubiv: Don't update then 😄

PR #348, opened 2020-10-14
Allow multiple modals

Joozty:


src/modules/modal/modal.md
```diff
@@ -161,9 +161,108 @@ const DemoThree = () => {
...
+      <Modal.Content tag="p">
+        When you click on <b>Close me!</b> button focus is returned back to <b>Modal 1</b><br/>
+        When you click on <b>Close all</b> button focus is returned back to initial button*.<br/>
+        
+        * Note that focus is not returned correcly because of
+        <a 
+            href="https://github.com/theKashey/react-focus-lock/issues/126"
+            target="_blank"
+            rel="noopener">
+                upstream bug
```
Joozty, line 219: This sadly hasn't been solved yet but I assume we can merge it regardless.  If the issue is not solved in foreseeable future I will try to come up with different solution.

viktor-yakubiv (commented):
I am not sure, multiple modal windows is a good idea, especially in a way it's implemented: even if I do so, I would prefer to overlap modals so the user doesn't lose the context.

The back button doesn't behave as it expected:
- if the user opens modal and clicks 'back' the modal should close
- if the user opens second modal and clicks 'back' the modal should close
- if the modal was closed confirming some action it should not be opened using a 'back' button 🤔 I am not sure in this

Why do you need multiple modals?

Joozty:
> Why do you need multiple modals?

This is needed for claiming on data provider page

> The back button doesn't behave as it expected

Do you mean the back button on the keyboard?


> I would prefer to overlap modals so the user doesn't lose the context.
Isn't this implementation exactly that?


Overall this is low-level implementation. It's up to the user how they handle the multiple modals... This PR just adds support for it.

viktor-yakubiv:
> Isn't this implementation exactly that?

I might not have noticed because modals are the same size.

viktor-yakubiv (approved):

PR #349, opened 2020-10-15
Out-of-the-box loading states 🧪

viktor-yakubiv:
Currently in the state of experiment. However, I think this will become production-ready very soon ✨

### DIFF
--- a/src/foundation/README.md
+++ b/src/foundation/README.md
@@ -42,3 +42,65 @@ to a parent element.
   <h6>Heading 6</h6>
 </div>
 ```
+
+## Loading states 🧪
+
+**Experimental** loading states.
+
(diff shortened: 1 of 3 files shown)


viktor-yakubiv:
Hmm, an image placeholder is an interesting problem. Technically, if render `img:not([src])`, an onerror will be thrown. However, practically it looks good.

viktor-yakubiv:
I decided:
1. Don't cover inline elements: there are 2 many of them and it's hard to do it properly, better if done in place
2. Cover image elements (added)
3. Leave the technique experimental

Joozty (approved):
I was really looking forward to this PR. You have done a tremendous job. 👍

PR #350, opened 2020-10-19
SearchResult component

viktor-yakubiv:
The API is more or less good but the feedback would be appreciated 🙂

## TODO

1. Loading state after #349 
1. Better name for the component?
2. Proper structuring?

### DIFF
--- a/src/modules/search-result/author.jsx
+++ b/src/modules/search-result/author.jsx
@@ -1,0 +1,35 @@
+import React from 'react'
+
+import styles from './author.css'
+
+import { classNames } from 'utils'
+
+const List = ({ children, className, ...htmlProps }) => (
+  <ol className={classNames.use(styles.list).join(className)} {...htmlProps}>
+    {children}
+  </ol>
+)
+
+const Item = ({ children, href, ...htmlProps }) => {
+  const ElementType = href != null ? 'a' : 'span'
+  return (
+    <li
+      itemProp="author"
+      itemScope=""
+      itemType="https://schema.org/Person"
+      {...htmlProps}
+    >
+      <ElementType href={href} itemprop="name">
+        {children}
+      </ElementType>
+    </li>
+  )
+}
+
+List.displayName = 'AuthorList'
+Item.displayName = 'AuthorList.Item'
+
+List.Item = Item
+
+export default List
+export { List, Item }
--- a/src/modules/search-result/search-result.jsx
+++ b/src/modules/search-result/search-result.jsx
@@ -1,0 +1,133 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import Metadata from './metadata'
+import AuthorList from './author'
+import styles from './styles.css'
+
+import { Card, Heading, Link } from 'elements'
+import { classNames } from 'utils'
+
+const texts = {
+  pdfAvailable: 'Get PDF',
+  linkAvailable: 'Access PDF',
+  metadataOnly: 'No full text',
+}
+
+const PREFIX = 'work'
+
+const labels = {
+  author: 'Author',
+  fieldOfStudy: 'Field of study',
+  publicationDate: 'Publication date',
+  publicationVenue: 'Publication venue',
+}
+
+const formatDate = (date) => date?.toString()
+
+const SearchResult = ({
+  children,
+  id,
+  className,
+  data: {
+    title,
+    author = [],
+    publicationDate,
+    publicationVenue,
+    fieldOfStudy,
+    metadataLink,
+    fullTextLink,
+    thumbnailUrl,
+  } = {},
+  ...htmlProps
+}) => {
+  const idFor = (scope) => `${id}-${scope}`
+  return (
+    <Card
+      id={`${PREFIX}-${id}`}
+      className={classNames.use(styles.container).join(className)}
+      itemScope
+      itemType="https://schema.org/ScholarlyArticle"
+      {...htmlProps}
+    >
+      <Heading level="3" className={styles.title} itemProp="name">
+        {metadataLink ? <Link href={metadataLink}>{title}</Link> : title}
+      </Heading>
+
+      <Metadata>
+        <Metadata.Item id={idFor('author')} label={labels.author}>
+          <AuthorList aria-label={labels.author}>
+            {author.map(({ name }) => (
+              <AuthorList.Item>{name}</AuthorList.Item>
+            ))}
+          </AuthorList>
+        </Metadata.Item>
+
+        <Metadata.Item id={idFor('venue')} label={labels.publicationVenue}>
+          {publicationVenue}
+        </Metadata.Item>
+
+        <Metadata.Item
+          id={idFor('publication-date')}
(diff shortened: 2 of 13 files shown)


Joozty:
I don't think this should be a part of our design system, should it? I mean for testing the loading state it's good but I would suggest to move it to search afterwards. What do you think?

Overall it looks good to me. You haven't covered some edge cases like the title too long, more authors yet. I don't see it in the TODO so I am just reminding you. 

Thanks 🙂

viktor-yakubiv:
I think it should because this can be used within multiple projects: Search, Reader, Recommender and even static website.

viktor-yakubiv:
I've just noticed that I don't like the API very much.

src/modules/search-result/author.jsx
```diff
@@ -0,0 +1,35 @@
...
+  </ol>
+)
+
+const Item = ({ children, href, ...htmlProps }) => {
+  const ElementType = href != null ? 'a' : 'span'
+  return (
+    <li
+      itemProp="author"
+      itemScope=""
+      itemType="https://schema.org/Person"
```
Joozty, line 19: Maybe it would be better if we handle this in JSONLD. We already use everywhere else but I am leaving this decision up to you.

src/modules/search-result/search-result.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+import Metadata from './metadata'
+import AuthorList from './author'
+import styles from './styles.css'
+
+import { Card, Heading, Link } from 'elements'
+import { classNames } from 'utils'
+
+const texts = {
+  pdfAvaliable: `Get PDF`,
```
Joozty, line 12: typo 👮

src/modules/search-result/search-result.jsx
```diff
@@ -0,0 +1,132 @@
...
+  return (
+    <Card
+      id={`${PREFIX}-${id}`}
+      className={classNames.use(styles.container).join(className)}
+      itemScope
+      itemType="https://schema.org/ScholarlyArticle"
+      {...htmlProps}
+    >
+      <Heading level="3" className={styles.title} itemProp="name">
+        {metadataLink ? <Link href={metadataLink}>{title}</Link> : title}
```
Joozty, line 54: What about long title? Shouldn't it be clipped?

src/modules/search-result/search-result.jsx
```diff
@@ -0,0 +1,132 @@
...
+      </Heading>
+
+      <Metadata>
+        <Metadata.Item id={idFor('author')} label={labels.author}>
+          <AuthorList aria-label={labels.author}>
```
Joozty, line 59: What about more authors? Maybe you could take inspiration in my PR? I've handled it there.

Joozty (commented):
It looks nice. It would be great if you could handle more authors and title. If you want to move it to the next iteration that's fine but I would suggest to not include it in the release till then, what do you think?

src/modules/search-result/author.jsx
viktor-yakubiv: Can you tell more how JSON-LD can be integrated in this the particular case?

styleguide.config.js
```diff
@@ -54,6 +54,10 @@ module.exports = {
       name: 'Modules',
       components: 'src/modules/!(select|modal)/**/*.{js,jsx,ts,tsx}',
       sections: [
+        {
+          name: 'Search result',
+          components: ['src/modules/search-result/search-result.jsx'],
```
Joozty, line 59: please exclude it here https://github.com/oacore/design/blob/master/styleguide.config.js#L55 So we don't have duplicate entries in menu

src/modules/search-result/search-result.jsx
viktor-yakubiv: I wanted to move this part to the next iteration. I am not sure how to properly do this from the semantic point of view.

Joozty: I would say my solution is semantically correct.

src/modules/search-result/author.jsx
Joozty: For example, I did integration here https://github.com/oacore/search/blob/master/pages/data-providers/index.jsx#L87. I don't know I find it easier to use. And I somewhere read that's it's recommended way for adding structured data to the website

viktor-yakubiv: I see this is quite nice. But I still don't understand how or where you can embed this if you have multiple instances of the same thing on the page or you propose to don't do it?

Joozty: That's something I don't know. 🤷‍♂️

Joozty (approved):
Feel free to merge after you solve conflicts and my open comments. Thanks for your work :)

src/modules/search-result/search-result.jsx
```diff
@@ -0,0 +1,132 @@
...
+            {author.map(({ name }) => (
+              <AuthorList.Item>{name}</AuthorList.Item>
```
Joozty, line 61: you should pass down id here

src/modules/search-result/search-result.md
```diff
@@ -0,0 +1,41 @@
...
+  <p data-line-count="2"></p>
+</SearchResult>
+```
+
+
+### Regular state
+
+```js
+<SearchResult
+  id="example-02"
```
Joozty, line 14: when I use article 82984538 the get pdf is not fully stretched
image

src/modules/search-result/search-result.jsx
viktor-yakubiv: Why? Which ID?

I made 2 ones: document ID that should be under `data.id` and the `id` attribute which is used in the DOM.

src/modules/search-result/search-result.md
viktor-yakubiv: That's terrible. I don't understand why this happens but I will have a look. Thank you.

viktor-yakubiv: It's weird, it doesn't happen to me and seems to be some cache problem.

The border is not applied to the container but to the image and even if `box-sizing: border-box` is not applied (but should, I forgot about it), the caption below just stretches to the full width.

In your example, the border goes around what makes me think something goes wrong.

viktor-yakubiv (commented):
@Joozty I need your answers to merge this. I finished all other comments at the moment.

src/modules/search-result/search-result.md
Joozty: Oh I am sorry I found out this issue while working on my [PR](https://github.com/oacore/design/pull/352). Unfortunately, I used the wrong target branch. It seems it was your previous branch and now you removed it or I didn't fetch new commits. I don't know. Sorry for confusing you. The issue is not present here.

src/modules/search-result/search-result.jsx
Joozty: The same as here https://github.com/oacore/design/pull/350

Joozty (approved):

PR #352, opened 2020-10-21
More authors

Joozty:
Video with screen reader enabled

https://www.loom.com/share/d1d39f8703814893a15d43c489197cd5

### DIFF
--- a/src/modules/search-result/author.jsx
+++ b/src/modules/search-result/author.jsx
@@ -1,28 +1,80 @@
-import React from 'react'
+import React, { useState, useEffect, useRef } from 'react'
 
 import styles from './author.css'
 
+import { Button } from 'elements'
 import { classNames } from 'utils'
 
-const List = ({ children, className, ...htmlProps }) => (
-  <ol className={classNames.use(styles.list).join(className)} {...htmlProps}>
-    {children}
-  </ol>
+const Item = ({ name, className, ...htmlProps }) => (
+  <li
+    itemProp="author"
+    itemScope=""
+    itemType="https://schema.org/Person"
+    className={classNames.use(className)}
+    {...htmlProps}
+  >
+    <a
+      href={`https://core.ac.uk/search?q=author:(${name})`}
+      className={styles.link}
+    >
+      {name.replace(',', ' ')}
+    </a>
+  </li>
 )
 
-const Item = ({ children, href, ...htmlProps }) => {
-  const ElementType = href != null ? 'a' : 'span'
+const List = ({ authors, className, ...htmlProps }) => {
+  const showMoreRef = useRef(null)
+  const [isExpanded, setIsExpanded] = useState(false)
+
+  useEffect(() => {
+    if (isExpanded)
+      showMoreRef.current.parentElement.nextElementSibling.firstChild.focus()
+  }, [isExpanded])
+
+  if (authors.length <= 4) {
+    return (
+      <ol
+        className={classNames.use(styles.list).join(className)}
+        {...htmlProps}
+      >
+        {authors.map((author) => (
+          <Item name={author.name} />
+        ))}
+      </ol>
+    )
+  }
+
   return (
-    <li
-      itemProp="author"
-      itemScope=""
-      itemType="https://schema.org/Person"
-      {...htmlProps}
-    >
-      <ElementType href={href} itemprop="name">
-        {children}
-      </ElementType>
-    </li>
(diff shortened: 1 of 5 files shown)


src/modules/search-result/author.jsx
```diff
@@ -1,28 +1,80 @@
...
-      itemScope=""
-      itemType="https://schema.org/Person"
-      {...htmlProps}
-    >
-      <ElementType href={href} itemprop="name">
-        {children}
-      </ElementType>
-    </li>
+    <ol className={classNames.use(styles.list).join(className)} {...htmlProps}>
+      <Item aria-setsize={authors.length} name={authors[0].name} />
```
viktor-yakubiv, line 49: You did not use the `aria-posinset`, it's recommended by the spec to combine these 2.

src/modules/search-result/author.jsx
```diff
@@ -1,28 +1,80 @@
...
+        className={classNames.use(styles.showMore, isExpanded && styles.hide)}
+      >
+        <Button
+          ref={showMoreRef}
+          onFocus={() => {
+            setIsExpanded(true)
+          }}
+          title="Show more authors"
+        >
+          + {authors.length - 3} MORE
```
viktor-yakubiv, line 62:
```suggestion
          +{authors.length - 3} more
```

Change the case using `text-transform: uppercase` but keep the text always in the correct case. I would call it design's backward compatibility.

If you want to add a space after `+` use some thinner one, a thin or a hair would be good, or up to your eye. Use [Unicode spaces](http://jkorpela.fi/chars/spaces.html) as a reference.

src/modules/search-result/author.jsx
```diff
@@ -1,28 +1,80 @@
...
 
-const Item = ({ children, href, ...htmlProps }) => {
-  const ElementType = href != null ? 'a' : 'span'
+const List = ({ authors, className, ...htmlProps }) => {
+  const showMoreRef = useRef(null)
+  const [isExpanded, setIsExpanded] = useState(false)
+
+  useEffect(() => {
+    if (isExpanded)
+      showMoreRef.current.parentElement.nextElementSibling.firstChild.focus()
```
viktor-yakubiv, line 31: I am afraid this effect will work before the actual click. Let me check how exactly it works... ⏳

viktor-yakubiv (approved):
I like what I've seen, I like how it's implemented. It's very good 👍

There are some remarks in the comment that are easier to fix. Maybe I will improve this in the future after some checks but this is nice now.

Thank you!

src/modules/search-result/author.jsx
Joozty: there is `useLayoutEffect` which I wanted initially use but then it worked even with useEffect so I went with it.  (`useLayoutEffect` is just client side hook and I didn't want to break the SSR)

src/modules/search-result/metadata.css
```diff
@@ -6,7 +6,7 @@
 .group {
   display: inline-block;
 
-  &:not(:last-child)::after {
+  &:not(:first-child)::before {
```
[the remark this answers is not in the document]
Joozty: I wanted avoid this empty line:

::after
after
::before
before

src/modules/search-result/author.jsx
Joozty: Isn't it only required when I want to have a different order of items?

src/modules/search-result/author.jsx
viktor-yakubiv: Yes, it does. Once I mouse-down it, the list changes but the click should happen only when the mouse is released.

It's not about the `useEffect` hook, it's the way it's implemented.

Joozty: Alright but I don't think this is a big deal, is it?

src/modules/search-result/metadata.css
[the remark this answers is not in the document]
viktor-yakubiv: I think you can avoid it converting the list to `inline`. In my opinion, it's the right way.

src/modules/search-result/author.jsx
viktor-yakubiv: It's not 🙂

src/modules/search-result/metadata.css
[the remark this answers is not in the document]
Joozty: Not really it didn't work for me. If you want you can try it. However I will merge this one, feel free to improve it revert it back if you find a solution for it.

PR #353, opened 2020-10-21
Update colour system to be more powerful and flexible

viktor-yakubiv:
1. Introduces a new more powerful and flexible technique of using colours and CSS custom properties. The system requires a lot of changes in the project, the slow adaptation approach is taken to replace the --primary colour.
2. Defines the new colour variables that are inspired by the [Material Baseline Design Kit](https://www.figma.com/community/file/778763161265841481/Material-Baseline-Design-Kit) developed by the Material Design group and provided via Figma Community.
3. Deprecates the current colour system in favour of the new more simple one.

This comes from stale #345. Merging can be either proceed or postponed to the time I test the system.

### DIFF
--- a/src/foundation/colors.css
+++ b/src/foundation/colors.css
@@ -3,12 +3,21 @@
  *
  * Design system colour palette. Contains shades of grey and our brand
  * colours.
+ *
+ * Guidelines:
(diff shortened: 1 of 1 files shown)


viktor-yakubiv:
This is properly tested and working!

Joozty (approved):

PR #356, opened 2020-10-26
Improve microdata and add tag as a prop

Joozty:


src/modules/search-result/search-result.jsx
```diff
@@ -29,6 +29,7 @@ const SearchResult = ({
   children,
   id,
   className,
+  tag: Tag = Card,
```
viktor-yakubiv, line 32: I don't think this should work so. You can support tag but it's better to pass it forward to the Card.

What is your use case for this?

src/modules/search-result/search-result.jsx
```diff
@@ -74,7 +75,7 @@ const SearchResult = ({
         <Metadata.Item
           id={idFor('field')}
           label={labels.fieldOfStudy}
-          itemProp="dataPublished"
+          itemProp="about"
```
viktor-yakubiv, line 78: Oops, my mistake.

src/modules/search-result/search-result.jsx
Joozty: I have a component [Search.Result](https://github.com/oacore/search/blob/master/modules/search-layout/result.jsx) and I need to put SearchResult from here to it because it applies all grid styles I need otherwise I would need to import css styles modules and that's not good too.

viktor-yakubiv: I don't get it. If you put this SearchResult as a direct child and add one more class from your component, it would work, wouldn't it?

Joozty: yes it would but I don't see anything wrong doing it with Tag. It's one node less.

viktor-yakubiv: Card component in the root of the module is a part of the design that is why I don't like the idea. The component should not become something else than a Card when the element type itself could be changed.

Joozty: so basically if I keep tag in `....htmlProps` and pass it down to Card it's ok with you?

viktor-yakubiv: Yes, I considered such an option when wrote the component. You can change the actually rendered component/element but card classes will be always passed.

src/modules/search-result/index.js
```diff
@@ -1 +1 @@
-export SearchResult from './search-result'
+export default from './search-result'
```
Joozty, line 1: Honestly, I don't know why but without this change the import `import {SearchResult} from '@oacore/design'` doesn't work.

viktor-yakubiv: The first one is named. I think `./modules/index` exports it in some specific way kind of `export SearchResult from '...'` what exports default which wasn't exported here where you fixed it.

src/modules/search-result/search-result.jsx
```diff
@@ -103,7 +103,9 @@ const SearchResult = ({
         )}
       </figure>
 
-      <div className={styles.content}>{children}</div>
+      <div className={styles.content} itemProp="abstract">
```
viktor-yakubiv, line 106: It will not always be abstract that's why it's not added. It can be either a text snipper or an abstract otherwise depending on the context.

viktor-yakubiv (approved):

PR #358, opened 2020-10-28
Universal header

viktor-yakubiv:


viktor-yakubiv:
@Joozty I almost finished this. One ~small~ big thing left: develop and add the search field here.

src/modules/header/styles.css
```diff
@@ -0,0 +1,246 @@
...
+/* Special menus */
+.services-menu {
+  @media (--header-breakpoint) {
+    & .menu-list {
+      /* flex-direction: row; */
+      flex-wrap: wrap;
+      width: calc(3 * var(--header-menu-width));
+      max-height: 20rem;
+
+      /* &::after {
```
Joozty, line 210: this can be probably removed, cannot be?

src/modules/header/styles.css
```diff
@@ -0,0 +1,246 @@
...
+  margin: 0;
+  list-style: none;
+
+  &.expanded {
+    display: flex;
+  }
+
+  @media (--header-breakpoint) {
+    position: absolute;
+    top: 100%;
```
Joozty, line 143: this rule doesn't change anything, I think you can safely remove it

src/modules/header/styles.css
```diff
@@ -0,0 +1,246 @@
...
+  }
+}
+
+/* Special menus */
+.services-menu {
+  @media (--header-breakpoint) {
+    & .menu-list {
+      /* flex-direction: row; */
+      flex-wrap: wrap;
+      width: calc(3 * var(--header-menu-width));
```
Joozty, line 207: It may overflow. Maybe you should break it sooner?

image

Joozty (approved):
Nice work. Please have a look at my comments. I didn't find anything blocking. All the things can be done (if needed) later on.  So you got my approval. 🙂 


@viktor-yakubiv What do you think about hiding menu automatically on focus out? it could be done with `:focus, :focus-within` (`:hover`) rules somehow. As I user I find it quite useful don't you think? :)

src/elements/card/card.md
```diff
@@ -1,4 +1,5 @@
-```jsx
+```js
```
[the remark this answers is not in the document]
viktor-yakubiv: It's a leftover, I was playing with my broken syntax highlighting when some JSX appears in markdown code.

src/modules/header/styles.css
```diff
@@ -0,0 +1,246 @@
...
+    margin: -.25rem;
+    vertical-align: middle;
+    content: '';
+    background:
+      hsl(
+        var(--app-bar-color, var(--component-color)) /
+        var(--app-bar-disabled, var(--component-disabled))
+      );
+    transition: transform 50ms;
+    clip-path: path('M7.41,8.58L12,13.17L16.59,8.58L18,10L12,16L6,10L7.41,8.58Z');
```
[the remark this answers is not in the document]
viktor-yakubiv: It should look like an arrow down. Doesn't it?

src/modules/header/header.jsx
```diff
@@ -0,0 +1,177 @@
...
+    <AppBar
+      className={classNames.use(styles.container).join(className)}
+      tag="header"
+      {...passProps}
+    >
+      <AppBar.Item className={styles.logoContainer} href="/" tag="a">
+        <Logo className={styles.logo} />
+      </AppBar.Item>
+
+      <AppBar.Item className={styles.main}>{children}</AppBar.Item>
```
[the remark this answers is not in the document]
viktor-yakubiv: Why? Because it's not used? (perhaps, I don't remember)

src/modules/header/header.jsx
```diff
@@ -0,0 +1,177 @@
...
+        tag="nav"
+      >
+        <Menu
+          id="services"
+          label="Services"
+          className={styles.servicesMenu}
+          expanded={expandedMenu === 'services'}
+          onExpand={() => toggleMenu('services')}
+        >
+          <Item className={styles.servicesOverviewItem} href="/services">
```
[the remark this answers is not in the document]
viktor-yakubiv: If it gets integrated to oacore/next, it will break local development. I find such kind of URLs safe enough to go ahead. I didn't want to develop another complex solution for URL management here and kept it simple at the moment.

src/modules/header/styles.css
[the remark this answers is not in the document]
Joozty: it doesn't not for me. It looks like grey square because of this invalid rule.

src/modules/header/header.jsx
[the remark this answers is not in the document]
Joozty: yes `main` doesn't exist

src/modules/header/header.jsx
```diff
@@ -0,0 +1,177 @@
...
...3.06c-35.6 6.27-72.3.01-103.8-17.72a156.197 156.197 0 01-69-79.51l-438.9-1219.12 116.7-58.82 484.9 898.19c11 20.54 28.3 37 49.4 46.93 21.1 9.94 44.8 12.82 67.7 8.23l531.8-150.26" stroke="%23263238" stroke-miterlimit="10"/><path d="M10772.6 5060.75l-135.4 67.66" stroke="%23263238" stroke-width=".75" stroke-miterlimit="10"/><path d="M11810.9 4850.45s-74 74.06-37.1 129.53c21.8 32.04 38.5 67.32 49.3 104.54 0 0-30.4 74.06-6.1 74.06 24.4 0 12.2-55.47 55.5-42.97s86.3 86.25 49.4 110.94c-27.6 15.42-56.8 27.69-87.2 36.57a20.75 20.75 0 00-9.9 6.58c-2.6 3.11-4.2 6.9-4.6 10.92-.5 4.02.3 8.08 2.1 11.68 1.8 3.61 4.6 6.6 8.1 8.62 40.1 27.08 86.3 43.84 134.5 48.76 39.3.84 78.6-3.25 117-12.19 0 0 121.9-166.41 209.7-209.69 87.8-43.28 49.4-240.47 49.4-265.16 0-24.68-86.3-80.16-86.3-80.16s24.7-55.47-86.2-178.6c-111-123.13-320.7-148.12-326.8-104.84-2.1 11.46-.9 23.29 3.4 34.09a61.285 61.285 0 0021.3 26.87s-185-80.16-271.2 6.09c-86.3 86.25 215.7 284.36 215.7 284.36z" fill="%23263238" stroke="%23263238" stroke-miterlimit="10"/><path d="M11810.9 4850.45s150 182.87 372.2 200.55M11757.9 4906.23s69.8 202.98 358.4 260.89" stroke="%23fff" stroke-width=".5" stroke-linecap="round" stroke-linejoin="round"/></svg>`
```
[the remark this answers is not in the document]
viktor-yakubiv: It's oacore/design, built with pure Babel and PostCSS. Importing will require adding some stuff to the build system for a single image. It's not worth it

src/modules/header/header.jsx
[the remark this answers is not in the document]
Joozty: oh, that's true. What about basePath prop for header?

src/modules/header/styles.css
viktor-yakubiv: No, the menu is smaller than `768px`. The breakpoint is applied to the full width of the screen when the menu is visible only on part of it. It should be safe to leave it as is.

src/modules/header/header.jsx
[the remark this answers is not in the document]
viktor-yakubiv: Is it really needed? It may get confusing soon if we add `/data-providers` or `/works` or something related to the oacore/search. I am afraid there is no better way to handle this than it's now.

src/modules/header/header.jsx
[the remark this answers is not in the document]
viktor-yakubiv: You rely on some ID in Search for the portal search field. Should it be included here?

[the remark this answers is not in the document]
Joozty: shoot I forgot about it. The problem is not with the id, that can be solved differently but with applying [classes](https://github.com/oacore/search/blob/master/templates/data-providers/styles.module.css#L10) to the item. Cannot the interface be slightly changed, i.e. any children passed down to header will be rendered in the middle?


```
<Header>
      <AppBar|Header.Item><Select>here</Select></AppBar|Header.Item>
</Header>
```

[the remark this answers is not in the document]
viktor-yakubiv: I was afraid of that, because it may break the spacing there. However, it won't not bad, I think. Let's try it

src/modules/header/styles.css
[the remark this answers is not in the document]
viktor-yakubiv: I see. It works in Firefox Dev, Chrome Canary and Safari 14 but doesn't work for the stable Chrome. I will add a `@supports` rule for it.

src/modules/header/styles.css
```diff
@@ -0,0 +1,246 @@
...
+.menu-list {
+  display: none;
+  flex-direction: column;
+  padding: 0;
+  margin: 0;
+  list-style: none;
+
+  &.expanded {
+    display: flex;
+  }
```
viktor-yakubiv, line 139: This flex breaks with the SVG image in Safari.

src/modules/header/styles.css
viktor-yakubiv: It's safer to have it there. If it's not there, `auto` is set which is equal to regular content placement. It produces a similar result but may get broken if the markup changes.

src/modules/header/styles.css
viktor-yakubiv: Yes, it's a leftover, thank you 🙂

src/modules/header/styles.css
```diff
@@ -0,0 +1,246 @@
...
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  line-height: 1; /* FIXME: in another place */
+}
+
+.container li::before {
+  /* What the hell? */
```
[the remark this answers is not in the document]
viktor-yakubiv: I get this but I don't know where it comes from

li::before problem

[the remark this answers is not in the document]
viktor-yakubiv: That's it: https://github.com/csstools/sanitize.css/blob/master/sanitize.css#L123-L129

I don't know now, should I keep it and find another trick or remove it...

src/modules/header/header.jsx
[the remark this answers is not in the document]
viktor-yakubiv: I did it. Please, be aware, I am centring that element applying `margin-*: auto` to the logo and the nav.

src/modules/header/styles.css
```diff
@@ -0,0 +1,268 @@
...
+
+.toggle-button {
+  --icon-color: hsl(var(--component-color) / var(--component-medium-emphasis));
+  padding: .75rem;
+  margin: -1rem;
+  line-height: 1.5;
+  border-radius: 0;
+}
+
+.menu {
```
Joozty, line 76: I think the menu needs some z-index and background colour

image

src/modules/header/styles.css
```diff
@@ -0,0 +1,268 @@
+@custom-media --header-breakpoint (min-width: 768px);
+
+:root {
+  --header-menu-width: 12rem;
+}
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  line-height: 1; /* FIXME: in another place */
```
Joozty, line 10: This is something we should be probably getting rid off. I mean if I define a fixed height for the header the buttons are not stretched to the whole area.  And this height has to be defined because the main area should be moved down by that offset. (--app-bar-skip)

src/modules/header/header.jsx
```diff
@@ -0,0 +1,177 @@
...
+  const [expandedMenu, setExpandedMenu] = useState('')
+  const toggleMenu = useCallback((nextId) =>
+    setExpandedMenu((prevId) => (prevId !== nextId ? nextId : ''))
+  )
+
+  return (
+    <AppBar
+      className={classNames.use(styles.container).join(className)}
+      tag="header"
+      {...passProps}
```
Joozty, line 94: please include `id`  here too :)

src/modules/header/styles.css
viktor-yakubiv: What do you mean saying 'get rid off'?

src/modules/header/styles.css
viktor-yakubiv: There is undocumented `--header-zindex` (or 1000 by default) applied for the desktop only. I will move this for mobile too.

viktor-yakubiv: I see the problem: if `position: fixed` is applied to the overall component, you lack the background and z-index. I am afraid, `z-index` should be applied manually.

viktor-yakubiv: Also, you can wonderfully drop `fixed` prop there which gets automatically passed to the `AppBar` and solves your problem.

viktor-yakubiv:
@Joozty I guess you had problems because I didn't push everything. Would you like to check it again? If not, I can rebase and merge this one, and we keep going

src/modules/header/styles.css
Joozty: I cannot drop position fixed because when you are on the mobile menu and open the submenu the main content gets moved down and I don't think it should work like that

src/modules/header/styles.css
Joozty: I mean if we have it there than buttons are not going to be stretched properly when you specify a height for app bar and imho it should by default

image

src/modules/header/styles.css
viktor-yakubiv: > There is undocumented `--header-zindex` (or 1000 by default) applied for the desktop only. I will move this for mobile too.

Renamed to `--header-z-index` and defaults to `--app-bar-z-index + 10`.

---

I don't know, perhaps not. But if you make the app-bar fixed, it won't do so, child menus will overlay the content of the page. I think it's good option.

src/modules/header/styles.css
viktor-yakubiv: Height for the AppBar? Why do you need it? I was trying to respect spacing as much as possible. The AppBar should be rendered in a good shape without height specified.

Joozty: That's how our app structure works in search as well as in dashboard (not relevant). And the reasons I can think of only one at the moment and I already described it here https://github.com/oacore/design/pull/358#discussion_r524409693 but there were probably more reasons?

Joozty: @viktor-yakubiv Any update on this? What's your decision here? I see this as the last blocker for merging this PR.

viktor-yakubiv: I finally got what's your problem is. Sorry, I did not get it before. This is how it looks for me:

Screenshot 2020-11-23 at 11 15 54

I will have a look at what I can do with `position: relative`

viktor-yakubiv: Please, have a look at the [solution](https://github.com/oacore/design/pull/358/commits/326c2f95a255b7cb6d784aba5f1cfc7681a6cb3d). Doesn't it feel bug-causing?

Joozty: I hope it's ok. Please remove `FIXME` comment, because it will stay there forever. 😄 and after that feel free to merge. Thanks a lot for your work :)

PR #360, opened 2020-10-30
Add custom BUILD_PATH support 🚧

viktor-yakubiv:
I found that if `npm link` doesn't work with Next.js it's possible to build the library to the sibling project's `node_modules` directory directly.

This PR introduces a way to make this work in an easy way:
- if `BUILD_PATH` environment variable exists, the project will build to that path
- otherwise, it will build to the current directory

This workaround doesn't work seamlessly because Next.js doesn't listen to changes in the `node_modules` but it's better than nothing.

### Example

```
export BUILD_PATH='/home/user/dev/oacore/next/node_modules/@oacore/design/'
npm run build
```

**NOTE:** Trailing slash is important to concatenate path correctly.

### Warnings

1. This is quite dangerous. I am not sure we want to support it.
2. I am not sure `npm link` doesn't work because I didn't test it properly.


Joozty:
I consider this as a big antipattern. On the other hand, because npm link doesn't work I often do it manually like this. I build the library and then copy-paste it to the sibling project. I am leaving up the merging decision on you.

package.json
```diff
@@ -83,10 +83,10 @@
     "format": "eslint --ignore-path .gitignore --ext .js,.jsx --fix .",
     "clean": "if [ -d lib ]; then rm -r lib; fi",
     "prebuild": "npm run clean",
-    "build:js": "babel src --out-dir lib --verbose",
+    "build:js": "babel src --out-dir \"${BUILD_PATH}lib\" --verbose",
     "build:css": "run-p 'build:css:* -- {@}' --",
-    "build:css:local": "postcss 'src/**/*.css' --base src --dir lib --verbose",
-    "build:css:global": "postcss 'src/**/*.css' --base src --env=global --dir lib/global --verbose",
+    "build:css:local": "postcss 'src/**/*.css' --base src --dir \"${BUILD_PATH}lib\" --verbose",
```
viktor-yakubiv, line 88: @Joozty I can make this a bit safer disabling it in production but it looks ugly and too hard to understand, in my opinion.

```suggestion
    "build:css:local": "postcss 'src/**/*.css' --base src --dir \"$( [ \\\"$NODE_ENV\\\" = 'production' ] && echo '' || echo $BUILD_PATH )lib\" --verbose",
```

Joozty: Rather not, it's really ugly 😄

viktor-yakubiv:
@Joozty I decided to merge this after #386

PR #386, opened 2020-11-03
Bump packages

Joozty:


Joozty:
I decided to downgrade postcss-calc because otherwise the css couldn't be imported in our projects. There was some **unknown error** from postcss-loader which made it even harder to debug.  I don't think postcss-calc is real cause of the issue. We should wait for [`postcss-preset-env`](https://github.com/csstools/postcss-preset-env/issues/191) to be upgraded and then we should try it again.

PR #388, opened 2020-11-16
Load result images lazily

Joozty:
Maybe it's just me. It looks to me that loading multiple images is slow. Maybe this could help our servers?


viktor-yakubiv (approved):
Thank you, I've never heard about this attribute. It's a year in the spec already. Thank you for bringing this to attention 👍 🙂

PR #389, opened 2020-11-17
Add search bar in Header

Joozty:
@viktor-yakubiv This is my attempt to integrate search bar in `<Header />`. After some thinking, I couldn't find a way how to do it via props as we discussed earlier so I went for the same approach as I did in @oacore/search. Most of the code is copy & paste

The main change is that I created a global context for the whole library. The interaction with the context is done via reducer similarly as we have it in the reader and the interface is very similar to redux one. So if we decided to go for it from now on it would be required to wrap whole application to `<DesignProvider />` in order to use the search bar. The reason for introducing global context rather than local one is that we may reuse it for our future components (or for example Modal). Let me know your thoughts 🙂 

TODO: Write documentation and do more testing


src/foundation/layout.css
```diff
@@ -0,0 +1,5 @@
+:root {
+  --page-padding-x: 6rem;
+  --page-padding-y: 3rem;
+  --main-content-percentage-width: calc(2 / 3);
+}
```
Joozty, line 5: These variables are required in order to have the search box the same width as the main content.

.github/workflows/pull-request-check.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         name: Setup Node
         uses: actions/setup-node@v2-beta
         with:
-          node-version: 12
+          node-version: 14
```
Joozty, line 20: I kept 12 everywhere because this is the only point for us where we can test that our projects will run on our servers. I no longer use this version and I don't check it against it. I always believe if something is wrong the GitHub actions will notify us.

Joozty: design and configs are probably safe to go but I wouldn't do it in other repos just yet. I am just letting you know my thinking behind upgrading things.

src/modules/header/styles.css
```diff
@@ -266,3 +268,70 @@
...
+
+  .select:focus-within {
+    position: absolute;
+    top: var(--app-bar-padding-y);
+    right: .5rem;
+    bottom: var(--app-bar-padding-y);
+    left: .5rem;
+    max-width: calc(100vw - var(--page-padding-x));
+  }
+}
```
viktor-yakubiv, line 337: Is it possible to rewrite this using a mobile-first approach? Also, joining media queries with the selectors themselves, i.e. incapsulating queries into selectors so they are not repeated?

styleguide.config.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
     {
       name: 'Modules',
       components:
-        'src/modules/!(modal|select|search-result)/**/*.{js,jsx,ts,tsx}',
+        'src/modules/!(modal|select|search-result|header)/**/*.{js,jsx,ts,tsx}',
```
viktor-yakubiv, line 56: Oh, did I forget about? Sorry

src/modules/header/styles.css
```diff
@@ -266,3 +268,70 @@
   font-size: .875rem;
   color: hsl(var(--app-bar-color, var(--component-color) / var(--component-medium-emphasis)));
 }
+
+
+/* search item styles */
+.search-bar-item {
+  --app-bar-padding-y: .25rem;
```
viktor-yakubiv, line 275: This spacing is pain here. As well as the logo, I thought I fixed it but I didn't, apparently. It's hard to think about how this could be fixed.

src/context.jsx
```diff
@@ -0,0 +1,40 @@
...
+  header: headerInitialState,
+}
+
+export const DesignContext = React.createContext(null)
+
+const reducer = ({ searchBar }, action) => ({
+  header: headerReducer(searchBar, action),
+})
+
+export const useDesignContext = () => {
```
viktor-yakubiv, line 15: I am not a supporter of global context but it's not bad either 😐

viktor-yakubiv (commented):
It's looking good to me 👍 It would be even better if the search input changes its look not on focus but on change, similarly to Chrome's address bar.

Also, proper vertical alignment and spacing is a headache here. Despite it breaks my soul of a perfectionist, I am considering this minor. If there is no way to handle vertical spacing, let's leave it.

Good job! 👍

.github/workflows/pull-request-check.yml
viktor-yakubiv: Thank you for letting me know.

viktor-yakubiv:
@Joozty sorry, I messed up committing to the wrong branch. I think the last commit should disappear when you rebase after merging #358.

src/modules/header/context/index.js
```diff
@@ -0,0 +1,6 @@
+export headerReducer from './reducer'
+export HEADER_ACTIONS from './actions'
```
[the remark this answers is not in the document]
Joozty: true however I didn't want to introduce any other dependency

src/modules/header/search-bar.jsx
```diff
@@ -0,0 +1,99 @@
...
+    </AppBar.Item>
+  )
+}
+
+const SearchBarOrEmpty = () => {
+  const searchBarConfig = useSearchBar()
+
+  if (!searchBarConfig) return null
+  return <SearchBar />
+}
```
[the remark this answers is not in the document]
Joozty: Not really, I would need to handle somehow conditional hooks. I will consider t it later but won't include it in this here. Thanks for the suggestion :)

viktor-yakubiv (approved):

src/modules/header/styles.css
```diff
@@ -52,6 +52,7 @@
 }
 
 .logo {
+  max-width: var(--page-padding-x);
```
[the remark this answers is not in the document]
Joozty: sadly yes, do you have any idea how to improve it?

src/modules/header/header.md
```diff
@@ -1,5 +1,60 @@
...
+    Header.useSearchBar({
+        onQueryChanged: () => {},
+        initQuery: '',
+        suggestionsDelay: 0,
+        searchBarProps: {
+          label: 'Search',
+          placeholder: 'e.g. repository or journal name',
+          prependIcon: '#magnify',
+        },
+        getSuggestions: (term) => options.slice(0, Math.max(0, options.length - term.length))
```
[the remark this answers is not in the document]
Joozty: then the options are not rendered

src/modules/header/styles.css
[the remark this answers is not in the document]
viktor-yakubiv: Maybe leave it as it. When I have final layouts, I will pick it up... Probably, forgetting already about this 😄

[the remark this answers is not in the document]
Joozty: I will remind you 😄

PR #394, opened 2020-11-30
Add ExpandableList

Joozty:


Joozty:
@viktor-yakubiv don't wanna push you but if you find spare time can you review with priority? Thx :)

src/modules/search-result/search-result.jsx
```diff
@@ -55,7 +54,22 @@ const SearchResult = ({
 
       <MetadataList>
         <MetadataList.Item id={idFor('author')} label={labels.author}>
-          <AuthorList aria-label={labels.author} authors={author} />
+          <ExpandableList aria-label={labels.author}>
+            {author.map((a) => (
+              <ExpandableList.Item
+                itemProp="author"
+                itemScope=""
+                itemType="https://schema.org/Person"
```
viktor-yakubiv, line 62: I am not sure whether the scope should be here or on the link below. I imagine item-props on the lowest level where only text is a child. However, it's my personal assumption, no reason for it behind.

src/elements/expandable-list/expandable-list.jsx
```diff
@@ -0,0 +1,83 @@
...
+import styles from './expandable-list.css'
+
+import { Button } from 'elements'
+import { classNames } from 'utils'
+
+const ExpandableListItem = ({ children, ...htmlProps }) => (
+  <li {...htmlProps}>{children}</li>
+)
+
+const ExpandableList = ({ children, className, ...htmlProps }) => {
```
viktor-yakubiv, line 12: I think it's too early to generalise the name because the component may become used inappropriately and difficult to understand what it is. Currently, this component is only the author list, it expands in a very specific way and displays in a very specific way too.

I find [these guidelines](https://namingthings.donedone.com) very useful to pick up good naming practices.

src/elements/expandable-list/expandable-list.jsx
```diff
@@ -0,0 +1,83 @@
...
+        className={classNames.use(styles.showMore, isExpanded && styles.hide)}
+      >
+        <Button
+          ref={showMoreRef}
+          onFocus={() => {
+            setIsExpanded(true)
+          }}
+          title="Show more"
+        >
+          + {childrenArray.length - 3} more
```
viktor-yakubiv, line 59:
```suggestion
          +{childrenArray.length - 3}&nbsp;more
```

src/elements/expandable-list/expandable-list.jsx
```diff
@@ -0,0 +1,83 @@
...
+      })}
+      <li
+        aria-setsize={childrenArray.length}
+        className={classNames.use(styles.showMore, isExpanded && styles.hide)}
+      >
+        <Button
+          ref={showMoreRef}
+          onFocus={() => {
+            setIsExpanded(true)
+          }}
```
viktor-yakubiv, line 56: For some reason, this doesn't work in Safari despite it works in Chrome. Perhaps, the button doesn't get focus on click and maybe `onClick` is needed too.

I would add `useCallback` here if you go to add `onClick` too.

viktor-yakubiv (approved):

src/modules/search-result/search-result.jsx
Joozty: I don't know. You may be right. I haven't really studied the microdata. Probably it's worth to put it in some validator 🤔

viktor-yakubiv: I could not find anything on the schema.org docs. Let's keep it as it is now.

viktor-yakubiv: But I find, they along with Wikipedia recommend Google's [Rich Search Results testing](https://search.google.com/test/rich-results) tool.

src/elements/expandable-list/expandable-list.jsx
Joozty: I imagined generalising component even more once we need it. For example support, different separators, different `show more` possitions etc. Currently, it is just authors list as you said.

PR #415, opened 2020-12-07
Add Chip

Joozty:


src/elements/badge/badge.css
```diff
@@ -0,0 +1,15 @@
...
+
+.badge {
+  padding: 2px 4px;
+  font-size: .625rem;
+  line-height: .75rem;
+  color: var(--primary);
+  text-transform: uppercase;
+  letter-spacing: .015em;
+  border-radius: 4px;
+  box-shadow: var(--elevation-01);
```
Joozty, line 14: @viktor-yakubiv I added elevation here even it is not in Figma. It looks better to me when the background is white. What do you think? Leave it or remove it?

viktor-yakubiv: Elevation increases the visibility of the element but also it makes the element seem to be interactive. In this case, we want neither of these features.

viktor-yakubiv:
You could call it a "badge" but I imagine [badges](https://getbootstrap.com/docs/4.5/components/badge/) as a slightly different UI pattern, I would call it similar to Material's [chips](https://material.io/components/chips).

In the case of the work page, it was a simple [overline](https://material.io/design/typography/the-type-system.html#applying-the-type-scale) (scroll to _Captions and overline_ section to find more), where one part of it was highlighted but probably shouldn't have been 🤔

Joozty:
> In the case of the work page, it was a simple overline (scroll to Captions and overline section to find more), where one part of it was highlighted but probably shouldn't have been 🤔

So it was mean to be treated as one string?

viktor-yakubiv:
Perhaps a paragraph with a couple of spans. It could also be a list (`dl` might be even better) but I didn't expect to have a special component for it, it's a single place use.

Joozty:
I was thinking of several components this morning:


1. **Chip** decided to add it here. No big reason for it but it seemed to me this is the best place for it
2. **CopyableInput** (for API card) - Keep it or the moment at the search and maybe in the future move it to design
3. **Select** for a different version of papers - Haven't decided yet but probably will keep it at the moment in search


If you think the design is not  a good place for chip feel free to close the PR and I will move it to search otherwise please review & approve. Thanks :)

viktor-yakubiv:
`Chip` component should look differently, I would reuse the Material's style for it. I will close the PR for time being but keep the branch, I hope it's reusable.

PR #416, opened 2020-12-09
Search bar alignment

Joozty:


src/modules/header/styles.css
```diff
@@ -326,10 +330,7 @@
 @media (--header-breakpoint-up) {
   .select {
     position: relative;
-    max-width:
-      calc(
-        (100vw - 2 * var(--page-padding-x)) * var(--main-content-percentage-width)
-      );
+    max-width: calc((100vw - 1rem - 2 * var(--page-padding-x)) *var(--main-content-ratio));
```
Joozty, line 333: 1rem - grid gap in search

src/modules/header/styles.css
```diff
@@ -340,3 +341,9 @@
     left: unset;
   }
 }
+
+@media (max-width: 118rem) and (min-width: 768px) {
```
Joozty, line 345: 100rem (main content width) + 7rem + 7rem (min-logo-width) + 1rem + 1rem (main content padding)

Joozty:
@viktor-yakubiv what do you think should be the max-width of main content? I arbitrarily chose 100rem but maybe you have a better/resonable number

src/modules/header/header.jsx
```diff
@@ -121,8 +121,10 @@ const Header = ({ children, className, id, ...passProps }) => {
       tag="header"
       {...passProps}
     >
-      <AppBar.Item className={styles.logoContainer} href="/" tag="a">
-        <Logo className={styles.logo} />
+      <AppBar.Item className={styles.logoContainer}>
```
viktor-yakubiv, line 124: We have `AppBar.Brand` which got forgotten and I don't even remember what it does.

src/foundation/layout.css
```diff
@@ -1,5 +1,15 @@
 :root {
-  --page-padding-x: 6rem;
+  --main-content-max-width: 100rem;
```
viktor-yakubiv, line 2: It's 1600px, it's too much. After 800px text becomes hard to read because it's too wide.

src/foundation/layout.css
```diff
@@ -1,5 +1,15 @@
...
+  --page-padding-x: calc((100vw - var(--main-content-max-width)) / 2);
   --page-padding-y: 3rem;
-  --main-content-percentage-width: calc(2 / 3);
+  --main-content-ratio: calc(2 / 3);
+  --main-content-aside-gap: 1rem;
```
viktor-yakubiv, line 6: I would use 3rem but I am sure what this is.

src/modules/header/styles.css
```diff
@@ -285,7 +311,7 @@
   --form-control-color: var(--gray-700);
   max-width: 100%;
 
-  padding: calc(.25rem - 1px) .25rem .25rem;
+  padding: calc(.25rem - 1px) 0 .25rem;
```
viktor-yakubiv, line 314: No right border looks a bit weird on the design preview but I hope it's fine on the full-width website.

viktor-yakubiv (approved):

src/modules/header/header.jsx
Joozty: The brand should get deprecated or rewritten. It doesn't align correctly IMHO.

src/foundation/layout.css
Joozty: What do you suggest then? Something in between like 1200px? This value is quite important because it cannot be changed in the application. It has to be changed here and it applies to all subprojects.

src/foundation/layout.css
Joozty: You mean you are not sure?

It's column number 2

image

src/foundation/layout.css
viktor-yakubiv: 1200 is fine.

src/foundation/layout.css
viktor-yakubiv: If it's `grid-column-gap` let's double it.

src/modules/header/styles.css
Joozty: The border is there. The padding is removed from left and right so all calculations are easier and also it aligns with result card correctly.

src/foundation/layout.css
Joozty, line 6:
```suggestion
  --main-content-aside-gap: 2rem;
```

src/modules/header/styles.css
Joozty: I see what you meant by your comment. This is just an issue with the preview. It shouldn't happen in production

PR #426, opened 2020-12-21
Add publisher name

Joozty:
Also fixes microdata errors.

Note: It's quite important to put span to link otherwise microdata are not recognised correctly.

Tested here https://search.google.com/structured-data/testing-tool

### DIFF
--- a/src/modules/search-result/search-result.jsx
+++ b/src/modules/search-result/search-result.jsx
@@ -36,8 +36,10 @@ const SearchResult = ({
     metadataLink,
     fullTextLink,
     thumbnailUrl,
-    repositoryName,
-    repositoryLink,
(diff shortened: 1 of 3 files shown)


src/modules/search-result/search-result.jsx
```diff
@@ -120,6 +122,23 @@ const SearchResult = ({
...
+        {repositoryLink && (
+          <div
+            itemProp="publisher"
+            itemScope
+            itemType="https://schema.org/Organization"
+          >
+            <Link
+              href={repositoryLink}
+              className={styles.repositoryLink}
+              external
```
Joozty, line 135: Maybe link without the icon?

viktor-yakubiv (approved):

